PR #11, opened 2018-02-11
Enable TCP support, switch to it by default, and increase MTU to 65535

Arceliar:
Like it says in the title.

Using a large MTU cuts down on syscalls for I/O with the tun device, which is a bottleneck on some platforms.

There's UDP chunking machinery already in place, which would allow big packets to be split up and reassembled (without relying on the OS to fragment them), but had been disabled until now. That was switched on.

TCP support was added, with a syntax for specifying if a manually added peer should be connected to over udp or tcp. The nice thing about TCP is that it can be given a big MTU packet without chunking, and then the OS can figure out how to deal with MTU. This turns out to be *significantly* faster than doing it in Yggdrasil. The good folks at @tomeshnet were reporting ~400 Mbps switch speeds on a rock64 single board computer. That's ~400 in and ~400 out, over a single GbE port, so it's getting pretty close to line rate. I think it's pretty impressive for a $25 board.

The down sides to TCP are head-of-line blocking, bufferbloat, and TCP meltdown. S for, I haven't seen meltdown issues, but it needs further testing. Bufferbloat is partially mitigated by having Yggdrasil maintain a LIFO stack of packets to send, which helps both reduce latency and indirectly notify the end points of a connection that there's congestion (by reordering packets, which some sane TCP stacks take as a signal to begin throttling back).

There's no fix for the TCP head-of-line blocking problem. The plan is to eventually add [SCTP](https://en.wikipedia.org/wiki/Stream_Control_Transmission_Protocol) support eventually, which is a *much* better fit for this kind of application, but I'm waiting to see if Go [adds their own implementation](https://github.com/golang/go/issues/22191) before I try to do anything myself.

In the interest of testing the TCP stack, auto-peering with link-local addresses currently uses the TCP transport, and any manually added peers that don't explicitly specify a protocol (with `udp:` or `tcp:`) default to using TCP. These behaviors are subject to change, especially if/when SCTP makes it in.


neilalexander:
Just so that this is recorded somewhere, the EdgeRouter X saw an almost-four-fold increase in throughput with TCP+`bigmtu` when running iperf3 as an endpoint:
```
Connecting to host fd03:51e2:2f33:d3c1:ea1f:703a:79a5:7b01, port 5201
[  4] local fd00:1ace:f6d6:703b:8f2b:e638:6759:8dc1 port 52256 connected to fd03:51e2:2f33:d3c1:ea1f:703a:79a5:7b01 port 5201
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-1.00   sec  4.92 MBytes  41.1 Mbits/sec                  
[  4]   1.00-2.00   sec  3.75 MBytes  31.5 Mbits/sec                  
[  4]   2.00-3.00   sec  4.75 MBytes  39.9 Mbits/sec                  
[  4]   3.00-4.00   sec  4.31 MBytes  36.2 Mbits/sec                  
[  4]   4.00-5.00   sec  4.12 MBytes  34.6 Mbits/sec                  
[  4]   5.00-6.00   sec  4.37 MBytes  36.7 Mbits/sec                  
[  4]   6.00-7.00   sec  4.68 MBytes  39.3 Mbits/sec                  
[  4]   7.00-8.00   sec  4.44 MBytes  37.2 Mbits/sec                  
[  4]   8.00-9.00   sec  5.18 MBytes  43.5 Mbits/sec                  
[  4]   9.00-10.00  sec  4.43 MBytes  37.2 Mbits/sec                  
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-10.00  sec  45.0 MBytes  37.7 Mbits/sec                  sender
[  4]   0.00-10.00  sec  44.5 MBytes  37.3 Mbits/sec                  receiver
```
For comparison, the results before:
```
Connecting to host fd03:51e2:2f33:d3c1:ea1f:703a:79a5:7b01, port 5201
[  4] local fd01:3ade:8910:f7d6:c600:f55:93e:bf03 port 58332 connected to fd03:51e2:2f33:d3c1:ea1f:703a:79a5:7b01 port 5201
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-1.00   sec   957 KBytes  7.84 Mbits/sec                  
[  4]   1.00-2.00   sec  1004 KBytes  8.23 Mbits/sec                  
[  4]   2.00-3.00   sec  1.01 MBytes  8.50 Mbits/sec                  
[  4]   3.00-4.00   sec  1.01 MBytes  8.46 Mbits/sec                  
[  4]   4.00-5.00   sec  1.06 MBytes  8.90 Mbits/sec                  
[  4]   5.00-6.00   sec   951 KBytes  7.79 Mbits/sec                  
[  4]   6.00-7.00   sec  1.01 MBytes  8.44 Mbits/sec                  
[  4]   7.00-8.00   sec  1.10 MBytes  9.19 Mbits/sec                  
[  4]   8.00-9.00   sec  1.06 MBytes  8.85 Mbits/sec                  
[  4]   9.00-10.00  sec  1.07 MBytes  8.96 Mbits/sec                  
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-10.00  sec  10.2 MBytes  8.52 Mbits/sec                  sender
[  4]   0.00-10.00  sec  10.1 MBytes  8.49 Mbits/sec                  receiver
```
I suspect a lot of this is down to the fact that there are less context switches happening with TCP compared to UDP.

PR #23, opened 2018-03-01
Add support for running OpenBSD

neilalexander:
Adds support for OpenBSD.

This currently only supports TAP mode, as we don't currently have code to support the Protocol Information header (4 bytes) as this cannot be disabled in TUN mode on OpenBSD.

This code can also be used as a template for supporting other BSDs, although some of the structs and constants may have different orders or values, i.e. `struct tuninfo` is in a different order on FreeBSD.

### DIFF
--- a/misc/tests/tunbench-client.go
+++ b/misc/tests/tunbench-client.go
@@ -7,7 +7,7 @@ import (
 	"os/exec"
 	"time"
 
-	"github.com/songgao/water"
+	"github.com/neilalexander/water"
 )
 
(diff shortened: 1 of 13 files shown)


neilalexander:
This PR actually grew arms and legs a bit because it also includes sane TUN defaults per-platform, but on the plus side, this does mean that on OpenBSD using `-autoconf` works properly with `/dev/tap0`.

Also somehow I think the `admin functions` must have got merged in here at some point too.

PR #24, opened 2018-03-01
Add admin functions for adding peers and modifying TUN/TAP

neilalexander:
This adds new admin functions for adding peers and modifying the TUN/TAP interface without restarting Yggdrasil. Note that changes made in realtime are **not** written to configuration!

New commands are:

- `addPeer a.a.a.a:b` to add a peer using the default protocol
- `addPeer tcp:a.a.a.a:b` to add a TCP peer
- `addPeer udp:a.a.a.a:b` to add a UDP peer 
- `setTunTap none` to remove the TUN/TAP interface
- `setTunTap auto tun` to configure an unspecified TUN interface with default MTU
- `setTunTap auto tap 1500` to configure an unspecified TAP interface with MTU 1500
- `setTunTap /dev/tap0 tap 1280` to configure a specified TAP interface with MTU 1280

### DIFF
--- a/src/yggdrasil/tun.go
+++ b/src/yggdrasil/tun.go
@@ -66,7 +66,8 @@ func (tun *tunDevice) read() error {
 	for {
 		n, err := tun.iface.Read(buf)
 		if err != nil {
-			panic(err)
+			// panic(err)
+			return err
 		}
 		o := 0
 		if tun.iface.IsTAP() {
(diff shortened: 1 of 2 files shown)


src/yggdrasil/tun.go
```diff
@@ -66,7 +66,8 @@ func (tun *tunDevice) read() error {
 	for {
 		n, err := tun.iface.Read(buf)
 		if err != nil {
-			panic(err)
+			// panic(err)
+			return err
```
Arceliar: Note to self: don't just `go tun.read()`, since it returns an error. Go some function that catches the error, and can either restart the read if it's recoverable or, signal that there was a failure if it's not (log and/or panic).

PR #34, opened 2018-03-04
Tag CircleCI build artifacts with version and branch

neilalexander:
This versions and tags the build artifacts with branch names where appropriate.

The artifacts can be accessed for download at: `https://circleci.com/api/v1.1/project/github/yggdrasil-network/yggdrasil-go/latest/artifacts`


neilalexander:
The package versioning follows the following format: `w.x.yyyy`, where `w.x` is set in the `VERSION` file (ideally incremented during breaking changes) and `yyyy` is the build number generated from the number of git commits on the branch.

(`vyatta-yggdrasil` numbering also has an additional `-zzzz`, as the Vyatta package has other files that can be changed and therefore needs additional versioning.)

PR #49, opened 2018-03-06
Misc cleanup

Arceliar:
Some cleanup of files under the misc directory. None of the little go language test files were useful for anything, so rather than have them clutter things up I decided to remove them. We can always put them back later if we need them. The same goes for a few run scripts from old netns tests, and some unused python scripts from the sim folder.


Arceliar:
So it seems that PRs in CircleCI are given a branch name along the lines of `pull/49`, which was leading to problems with the build. Fixed by deleting `/` characters from the branch name before emitting the name used to build packages.

Issue #51, opened 2018-03-06
Different port numbers for TCP/UDP when port number not specified


neilalexander:
When using `-autoconf`, or alternatively when the `Listen` address is configured with a `:0` port number, the TCP and UDP sockets come up with different port numbers.

We can either write manual logic to randomly select the port number for both protocols (and extra logic to re-bind to a new random port if one or other protocol has that port number already in use), or we can leave the behaviour as-is and let the operating system assign unique ports for each protocol as it does today.

If the latter, both the TCP and the UDP listen addresses/ports should be written out to the console when Yggdrasil starts (whereas right now only the TCP port is).

Arceliar:
It probably makes sense for every listening socket to print sock.LocalAddr() to the log, whether or not we decide to add manual logic to fix the random ports. It's useful from a debugging perspective to know what address it binds to, in case it doesn't match what the user expects from the config (since Go seems to have some interesting ideas about what to listen on when given an unspecified address).

Arceliar:
So there's still the question of what to do when the listen address has a `:0` port.

My line of thinking: incoming manually configured connections (not from link-local auto-peering) are impractical with a random port number. The users need to change their config ever time something restarts, or manually add peers over the admin socket. So allowing TCP and UDP (and eventually SCTP?) to select different ports, in the case where none was specified, doesn't seem like it's hurting anything for now. If that's true, then not adding the extra checks is simpler than having them, so I'd lean towards keeping the current behavior, with the caveat that this could change (users shouldn't depend on TCP and UDP getting different ports, it's an unspecified behavior that happens to give different ports right now).

Another option would be to have ListenTCP and ListenUDP configuration options, in case someone wants e.g. fixed UDP (to manually configure nodes for UDP hole punching, if stuck behind a NAT outside of the user's control), but keep random TCP for whatever reason.

neilalexander:
I think I would agree - it is probably better to keep the current behaviour. 

Perhaps we can modify `-genconf` to generate a port number instead of inserting `:0` so that at least out of the box, it sets up expected configuration. If the port is in use then the user can just modify the configuration file. 

For `-autoconf` the current random port generation is fine because there is no expectation of persistence and the current method guarantees that autoconfiguration will always find unused ports. 

I think `ListenUDP` and `ListenTCP` are overkill.

Arceliar:
I think we can safely close this one, random different ports in the special case that `:0` was used is probably an OK behavior to keep.

PR #53, opened 2018-03-07
Generate a random port number in -genconf instead of :0

neilalexander:
Addresses concern in #51 by generating a random port number for `Listen` when running `-genconf`.

For `-autoconf` the listen address is left as `:0` to ensure that the operating system can always pick free ports for autoconfiguration, even if they are not matching.

### DIFF
--- a/yggdrasil.go
+++ b/yggdrasil.go
@@ -113,7 +113,7 @@ func generateConfig(isAutoconf bool) *nodeConfig {
 		cfg.Listen = "[::]:0"
 	} else {
 		r1 := rand.New(rand.NewSource(time.Now().UnixNano()))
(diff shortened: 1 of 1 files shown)


Arceliar:
I like it in principle, but maybe the random port should start higher. It would be good to eliminate any risk that the default random listen port will match the admin socket port or the UDP link-local multicast port (both of which are currently 9001). If not 1024, then I have no idea what a good starting point would be for a random port. 16384? 32768?

I guess we should also avoid the listen conflict in cases where `:0` is used, by opening the admin and multicast sockets before we try to open the TCP/UDP listen sockets, but that's a separate problem.

PR #74, opened 2018-05-07
Partial support for authenticated peers

Arceliar:
This is the backwards and (mostly) forwards compatible part of adding optional authentication to peers.

It adds a `PeerBoxPubs []string` to the config, and a `addAuthBoxPub` to the admin API. When connecting to a peer, the following rules are enforced:
1. Link-local IPv6 is always allowed, unless forbidden by other options in the config (multicast, linklocal).
2. All outgoing TCP (and `socks://...`) is always allowed.
3. Authorized incoming TCP is allowed.
4. Authorized incoming and outgoing UDP is allowed.
5. Other (unauthorized) links are not allowed.

If the list of authorized keys is empty (default), then all traffic is considered to be authorized (the current default behavior).

In addition, traffic from a peer is ignored until a successful switch announcement is received. If no switch announcement has been received for some time (currently 16 seconds), then a peer will close its connection. This partly mitigates attacks where an attacker observes authorized keys from the exchange and begins key exchanges to open connections and waste resources.

More elaborate replay attacks are still possible, such as by replaying one valid connection's link protocol traffic on top of multiple spoofed connections, which should cause switch announcements to get through. It would make sense to address this as part of the breaking changes for the next release.

### DIFF
--- a/src/yggdrasil/admin.go
+++ b/src/yggdrasil/admin.go
@@ -377,7 +377,7 @@ func (a *admin) getAuthBoxPubs() string {
 
 func (a *admin) addAuthBoxPub(bstr string) (err error) {
 	boxBytes, err := hex.DecodeString(bstr)
-	if err != nil {
+	if err == nil {
 		var box boxPubKey
 		copy(box[:], boxBytes)
 		a.core.peers.addAuthBoxPub(&box)
@@ -387,7 +387,7 @@ func (a *admin) addAuthBoxPub(bstr string) (err error) {
 
 func (a *admin) removeAuthBoxPub(bstr string) (err error) {
 	boxBytes, err := hex.DecodeString(bstr)
-	if err != nil {
+	if err == nil {
 		var box boxPubKey
 		copy(box[:], boxBytes)
 		a.core.peers.removeAuthBoxPub(&box)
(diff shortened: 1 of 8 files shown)


src/yggdrasil/config/config.go
```diff
@@ -5,6 +5,7 @@ type NodeConfig struct {
 	Listen      string
 	AdminListen string
 	Peers       []string
+	PeerBoxPubs []string
```
neilalexander: Internally the name `PeerBoxPubs` I guess is accurate, but it isn't an intuitive name, and someone looking at the nodeconfig without documentation would probably have no idea what this does. Would it be better to give it a more friendly name, i.e. `AllowedBoxPubs` or `AllowedPublicKeys` or so?

src/yggdrasil/admin.go
```diff
@@ -104,6 +105,23 @@ func (a *admin) init(c *Core, listenaddr string) {
 			*out = []byte(a.printInfos([]admin_nodeInfo{info}))
 		}
 	})
+	a.addHandler("getAuthBoxPubs", nil, func(out *[]byte, _ ...string) {
```
neilalexander: Same as `PeerBoxPubs` in nodeconfig - we might want to give this a slightly friendlier name.

Issue #75, opened 2018-05-07
Review configuration option naming


neilalexander:
Some of the configuration options in nodeconfig may not be very obvious and could do with some review.

Some obvious candidates for change:

- `BoxPub` and `BoxPriv`
- `SigPub` and `SigPriv`
- `LinkLocal`

PR #76, opened 2018-05-09
Use HJSON for config instead of JSON

neilalexander:
JSON does not permit comments and may not be the most user-friendly format for the Yggdrasil configuration file.

After exploring YAML, TOML and others, I propose a switch to HJSON. I have made some modifications to the [hjson-go](https://github.com/neilalexander/hjson-go) library to support inserting comments using `struct` tags, and this has also been submitted as a PR for upstream acceptance.

This PR modifies `-genconf` to produce HJSON commented output, and modifies `-useconf` and `-useconffile` to parse HJSON.

It so happens that HJSON is backward-compatible with JSON, therefore existing configuration files should continue to work in the meantime and have done in my limited testing (until some changes are made to address #75 that is).

### DIFF
--- a/src/yggdrasil/config/config.go
+++ b/src/yggdrasil/config/config.go
@@ -5,7 +5,7 @@ type NodeConfig struct {
 	Listen         string    `comment:"Listen address for peer connections (default is to listen for all\nconnections over IPv4 and IPv6)"`
 	AdminListen    string    `comment:"Listen address for admin connections (default is to listen only\nfor local connections)"`
 	Peers          []string  `comment:"List of connection strings for static peers (i.e. tcp://a.b.c.d:e)"`
-	AllowedBoxPubs []string  `comment:"List of peer BoxPubs to allow UDP incoming TCP connections from"`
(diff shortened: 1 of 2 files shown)


Arceliar:
I'm not 100% happy with hjson as a format for the long term, but since it's backwards compatible with json, allows comments, and is less finicky about commas, I like this a lot for now. We can revisit the issue of using something else (maybe yaml) if/when we a decent library with support for generating commented output appears.

Two very minor complaints.
1. This brings in a new external dependency. There's not much that can be done about it, short of forking `github.com/mitchellh/mapstructure` and moving it and `hjson-go` fork into the org. I'm a little paranoid about packages disappearing without warning, since that's happened to other projects that use git repos as packages. No action needed for now, but it's the sort of thing to consider doing something about before a 1.0 release, if we ever get to that stage.
2. I'm probably not the only person who's not very familiar with hjson, and I wouldn't recognize it on sight. It maybe makes sense to mention that the format is hjson at the beginning of the config, maybe referencing hjson.org, and we should update our documentation to use `config.hjson` or similar.

PR #82, opened 2018-05-18
Track TX/RX bytes over session and if MTU was adjusted, adjust MTU exchange behaviour, send session pings on TUN/TAP change

neilalexander:
This exposes three new fields to the `getSession` admin socket call:

- `bytesSent` shows the number of real traffic bytes sent over a session (not including overheads)
- `bytesRecvd` shows the number of real traffic bytes received over a session (not including overheads)
- `wasMTUFixed` shows if the MTU had to be adjusted to a lower value as a result of read errors


neilalexander:
Some further changes:

- Nodes can now exchange an MTU of `0`, which happens when one or other end has no TUN/TAP device (i.e. through `"IfName": "none"`)
- In the event that a session MTU is `0`, send back _ICMPv6 Destination Unreachable_ messages to the sender instead of wasting network bandwidth sending the traffic to a dead end
- If the TUN/TAP adapter MTU is changed, or the TUN/TAP is taken up or down using the admin socket, send session pings to notify active sessions of the changes

PR #86, opened 2018-05-21
Show IPv6 address and subnet on startup

neilalexander:
This is a very minor fix to expose the IPv6 address and subnet from `Core`, which allows both to be displayed on stdout at startup.

This is really just for user-friendliness.

Sample output:
```
2018/05/21 16:15:13 Your IPv6 address is fd00:baf1:6864:d945:1f44:b820:6bb5:98a7
2018/05/21 16:15:13 Your IPv6 subnet is fd80:baf1:6864:d945::/64
```


neilalexander:
Yes, good point. I'll create a separate PR for that later.

PR #91, opened 2018-05-24
Remove all DEBUG calls from Yggdrasil package, fix multicast, add Core API

neilalexander:
This change makes a number of changes to remove `DEBUG_` calls from `debug.go` being used in the `yggdrasil` package.

This has resulted in a number of new functions being exposed through the `Core` interface and godoc'd appropriately, and `yggdrasil.go` has now been updated to use those functions. 

It also fixes the multicast discovery, which currently would enable itself even when no regexes were specified.

I haven't removed `debug.go` but the simulator should be tested to make sure that I haven't broken it.


neilalexander:
This is going back to the drawing board.

PR #92, opened 2018-05-26
Use backpressure instead of estimated bandwidth

Arceliar:
Currently, routing uses some combination of (very poorly) estimated bandwidth and the estimated distance to the destination based on the tree.

This change uses distance on the tree + the size of a local queue (in packets, not bytes), to sort-of fake backpressure routing. It's more like local backpressure--queue information doesn't propagate between nodes--but it seems to cover the same use cases as the bandwidth tracking (while being less complex and seemingly more reliable in some initial tests).

The main advantage is that it can switch between links while under load. Currently, if a slow link is under load and a fast link comes up, it tends to keep using the slow link. With these changes, tends to quickly switch to using the fast link, based on my initial tests. Theoretically, the total available throughput should be the sum of all links, but the goodput may be slightly lower than the fastest link (I think using multiple paths causes a little more packet reordering, so the iperf tcp stream throttles down a little more than usual).

I've run some quick tests over tcp links, but it still needs testing on udp, and I need to think more about if there's a cheap/free way to communicate some information about local queue sizes without negatively affecting the traffic streams that aren't flooding.


neilalexander:
This seems like a nice simple alternative.

I figure that when a new link comes up, the local backpressure of that link will be zero, so any new link (regardless of whether it is faster or slower) will be temporarily preferred until the queues balance.

I also expect that two links that are close in speed will nearly load-balance, although not predictably. This might not be ideal if it will result in a lot of packet reordering. Should we be preferring one or other link in that case?

Arceliar:
The case where there are multiple interfaces *already* seems to prefer one, which wasn't intentional, and I'm not certain is the right way to go. Or at least, I'm not certain we want it to be preferring the interface that it currently is. It iterates over all the interfaces in the lookup table in order, and that order is defined by the port number.

I'm thinking maybe it makes sense to sort the lookup table by uptime. That would prefer long-lived nodes in the event that two are otherwise identical in terms of distance and queue size.

PR #94, opened 2018-05-27
hjson bugfixes

Arceliar:
When loading a config file through hjson, convert it to json and then load the json into the config struct. This correctly allows missing values to take on the defaults normally associated with genconf.

### DIFF
--- a/yggdrasil.go
+++ b/yggdrasil.go
@@ -123,7 +123,8 @@ func doGenconf() string {
 var pprof = flag.Bool("pprof", false, "Run pprof, see http://localhost:6060/debug/pprof/")
 var genconf = flag.Bool("genconf", false, "print a new config to stdout")
 var useconf = flag.Bool("useconf", false, "read config from stdin")
-var normaliseconf = flag.Bool("normaliseconf", false, "use in combination with -useconf, outputs your configuration normalised")
+var useconffile = flag.String("useconffile", "", "read config from specified file path")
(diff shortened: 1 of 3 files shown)


neilalexander:
The reason `useconffile` was added was because it isn’t possible to redirect to stdin when running as a Windows service. By removing it, we would break the ability to run as a service on Windows.

yggdrasil.go
```diff
@@ -150,6 +144,11 @@ func main() {
 		if err := hjson.Unmarshal(config, &dat); err != nil {
 			panic(err)
 		}
+		confJson, err := json.Marshal(dat)
```
neilalexander: Appears that this does successfully fix the problem where setting `"MulticastInterfaces"` to `[]` didn't work.

yggdrasil.go
```diff
@@ -134,14 +132,10 @@ func main() {
 	switch {
 	case *autoconf:
 		cfg = generateConfig(true)
-	case *useconffile != "" || *useconf:
```
neilalexander: The reason `useconffile` was added is because it isn’t possible to redirect to stdin when running as a Windows service. By removing it, we would break the ability to run as a service on Windows using a static configuration file.

Arceliar: Reverted this, but we should think about if there's a way to do this without useconffile. Maybe it can be wrapped in a script or something? I don't know much about how Windows services are configured/run.

PR #98, opened 2018-06-01
Admin "dot" style changes

Arceliar:
The dot graph is a little bland. It shows you the nodes you know about, but says nothing about why you know them.

This change styles nodes differently based on why we know them. Currently they're colored as follows: nodes that we only know about because we have an active session are red, nodes we have in the DHT are blue, peers are green, the self node is white, and anything missing but inferred (`?`) is shaded.

I have no idea if that's something we want. If it is, then I have no idea if these color choices are appropriate (we should make sure that the text is easy to read, including for people with abnormal color vision, and maybe introduce a shape difference too).


Arceliar:
Looks OK to me. We can always change it later if we can come up with a better color / label scheme.

PR #103, opened 2018-06-08
Use new switchMsg format

Arceliar:
Note that this PR is *not* backwards compatible, so it's requested for `develop`, and shouldn't hit `master` until the v0.2 release.

This is a whole mess of changes related to using a new switchMsg format.

Previously, we had 3 packet types related to the switch / spanning tree. There were announcements, hop requests, and hop responses. The announcements listed some basic info about a node's coords (including a sequence number and the number of hops to the root), and the node's peers would send hop requests and get back hop responses until they had all the information they needed. `wire_LinkProtocolTraffic` packets bypass the LIFO queue in `tcp.go`, to make sure they aren't dropped due to congestion, and to avoid races with later traffic packets.

That made sense when we wanted to send tiny UDP packets. After deprecating the old udp implementation, it no longer makes sense to keep the 3 kinds of messages. Instead, there's 1 `switchMsg`, which contains all the information. Instead of being sent periodically, a node sends one to their peers immediately in response to a locator change (either new root, new coords, or a new timestamp from the root).

Related changes include:
1. Peers immediately find a new parent if their current parent goes offline, when possible.
2. Some reworking of how peers are added to the DHT, to deal with not constantly getting spammed with traffic from peers.
3. Some DHT tuning. The part that periodically does a lookup of a bucket and pings the best node that's older than 1 minute now uses a variable threshold per node. This is used to ping new nodes frequently, and gradually throttle back to the old 1 minute cooldown. It should help nodes bootstrap faster. The rate at which it throttles back hasn't been tuned, so it may need further study.
4. `wire_Traffic` and `wire_ProtocolTraffic` packets have had their `TTL` fields removed--packets are no longer mutated during forwarding. This is because ordered reliable transport eliminates the race condition that previously would lead to routing loops. It should still be possible for a packet to follow a loop *once* if there are nodes in the loop that are updating coords while the packet is in flight.
5. Since there's no longer idle traffic for each peer, `tcp.go` sends an empty message through the TCP stream after 4 seconds without sending any other traffic. The read side of the stream continues to time out after 6 seconds of inactivity, as before. Both values are subject to tuning.

### DIFF
--- a/misc/sim/treesim.go
+++ b/misc/sim/treesim.go
@@ -162,7 +162,9 @@ func testPaths(store map[[32]byte]*Node) bool {
 			for here := source; here != dest; {
 				temp++
 				if temp > 4096 {
-					panic("Loop?")
+					fmt.Println("Loop?")
+					time.Sleep(time.Second)
+					return false
 				}
 				nextPort := here.core.DEBUG_switchLookup(coords)
 				// First check if "here" is accepting packets from the previous node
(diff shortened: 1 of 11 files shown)


Arceliar:
Not much point in having a breaking label until the eventual merge with `master`.

PR #129, opened 2018-06-15
Convert config from UTF-16 if a BOM is found

### LINKED ISSUE #119: Configuration file encoding issues on Windows
This problem has been especially apparent on the latest version of Windows 10.

If you run `yggdrasil.exe -genconf > Yggdrasil.conf` from a Command Prompt, the file is written and can be read by `yggdrasil.exe -useconffile` and `yggdrasil.exe -useconf`. 

If you follow the same steps in PowerShell instead of Command Prompt, the encoding of the file is different (Unicode-perhaps?) and then Yggdrasil can no longer interpret it. (At a closer look, it seems that Yggdrasil reads every other byte of the file as `0x00`, which trips up `hjson`).

Also if you edit the `yggdrasil.conf` file in Notepad and then save, once again, the encoding is transformed and Yggdrasil can no longer make sense of it (same symptoms as with PowerShell).

neilalexander:
So at some point recently, Windows 10 has now started converting things needlessly to UTF-16. This happens not only in PowerShell but even in Notepad!

Since Windows tends to prefix the conversion with a BOM, it's possible to detect easily, therefore the configuration is converted back down to UTF-8 before giving it to `hjson`. This prevents the panic at startup.

As a result, this pretty much fixes #119 in effect. 

Some extra code will need to be added for cases where the file might be UTF-16 but without a BOM, but I guess this is a lot less likely, and it removes the immediate issue for Windows users in the meantime.

### DIFF
--- a/yggdrasil.go
+++ b/yggdrasil.go
@@ -1,7 +1,8 @@
 package main
 
-import "encoding/json"
+import "bytes"
 import "encoding/hex"
+import "encoding/json"
 import "flag"
 import "fmt"
 import "io/ioutil"
@@ -12,6 +13,8 @@ import "time"
 import "regexp"
 import "math/rand"
 import "log"
+
+import "golang.org/x/text/encoding/unicode"
 
 import "yggdrasil"
 import "yggdrasil/config"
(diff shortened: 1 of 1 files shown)


yggdrasil.go
```diff
@@ -107,6 +110,19 @@ func main() {
 		if err != nil {
 			panic(err)
 		}
+		// If there's a byte order mark - which Windows 10 is now incredibly fond of
+		// throwing everywhere when it's converting things into UTF-16 for the hell
+		// of it - remove it and decode back down into UTF-8. This is necessary
+		// because hjson doesn't know what to do with UTF-16 and will panic
+		if bytes.Compare(config[0:2], []byte{0xFF, 0xFE}) == 0 ||
+		  bytes.Compare(config[0:2], []byte{0xFF, 0xFF}) == 0 {
```
cathugger: shouldn't it be `[]byte{0xFE, 0xFF}`?
https://en.wikipedia.org/wiki/Byte_order_mark#Byte_order_marks_by_encoding

neilalexander: You are indeed correct - that's a typo. I'll push a fix to the `develop` branch for merging into the next release.

PR #135, opened 2018-06-16
Fix typo for big-endian BOM

neilalexander:
This was a typo that would have affected detecting a BOM for big-endian UTF-16.

This didn't affect Windows as it seems to rather tightly stick to little-endian.

Thanks @cathugger for pointing this out!


Arceliar:
Good catch

PR #146, opened 2018-06-25
Local backpressure improvements

Arceliar:
These changes give a more realistic implementation of backpressure using only local information.

Switch lookup logic has been moved to a dedicated switch worker goroutine, which does basically 2 things:
1. When a packet comes in, it checks all idle peers, and sends the packet to the one that's the best next hop, if a valid next hop exists. If no next hop exists, the packet is buffered, with overflowing buffers dropping in FIFO order.
2. When a connection to a peer notifies the switch that it's idle (i.e. able to send something), the worker loops over the buffers and finds the smallest non-empty buffer for which the peer is a valid next hop, and sends a packet from the buffer (FIFO). If it can't find a packet, it adds the link to the idle set (to be used by 1).

Separate buffers are maintained for each session (`coords` + `handle`) or protocol traffic pairing (`coords` + `toKey` + `fromKey`). Since the smallest non-empty buffer is used in 2, my *hope* is that it will tend to send traffic from mostly-idle connections first, to give latency-sensitive small traffic streams a higher priority than bulk data transfers. Separate buffers per session means that different nodes no longer interfere with eachother, outside of the TCP socket buffers, so a simple FIFO buffer can be used, which improves throughput. The down side is that there's multiple traffic sources in the same session (e.g. a connection to a VPN with both TCP and UDP streams in it), then the TCP tends to cause that whole session to bufferbloat. Switching back to LIFO would be a possible workaround for that, but I'd like to look for alternatives first.

The changes generally work, but more testing is needed to see if there are additional performance implications that I haven't noticed yet.

### DIFF
--- a/misc/run-schannel-netns
+++ b/misc/run-schannel-netns
@@ -51,12 +51,12 @@ ip netns exec node4 ip link set lo up
 ip netns exec node5 ip link set lo up
 ip netns exec node6 ip link set lo up
 
-ip netns exec node1 ./run --autoconf --pprof &> /dev/null &
-ip netns exec node2 ./run --autoconf --pprof &> /dev/null &
-ip netns exec node3 ./run --autoconf --pprof &> /dev/null &
-ip netns exec node4 ./run --autoconf --pprof &> /dev/null &
-ip netns exec node5 ./run --autoconf --pprof &> /dev/null &
-ip netns exec node6 ./run --autoconf --pprof &> /dev/null &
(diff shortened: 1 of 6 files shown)


Arceliar:
I'm pretty confident that this is better than what we currently have, so I think it's ready to review/pull.

neilalexander:
I had the `backpressure` branch running on a box of mine for the last couple of days without any issues, and the changes seem to make sense.

PR #152, opened 2018-07-06
Limit backpressure resource consumption

Arceliar:
Previously, we kept an arbitrary amount of information in the backpressure buffers, but removed anything older than 25 ms. That time is both arbitrary and needs work tuning.

This patch set changes the following:
1. Track queue sizes in terms of bytes instead of number of packets.
2. Limit the total size of all queues to 4 MB. If the queue sizes get larger than this limit, the first packet from a random queue is dropped, with odds based on relative queue sizes, until the total drops below 4 MB. The 4 MB threshold is arbitrary, but I think it's a reasonable size for now. We could either make this configurable, or adjusted at runtime based on e.g. number of peers, at some later point.
3. When deciding which queue to send from, select the one that maximizes `<age of oldest packet>/<size of queue>`. Previously, it always sent from the smallest queue. This change continues to prioritize small streams of traffic, but prevents high bandwidth queues from blocking completely.

### DIFF
--- a/src/yggdrasil/switch.go
+++ b/src/yggdrasil/switch.go
@@ -605,21 +605,16 @@ type switch_buffers struct {
 }
 
 func (b *switch_buffers) cleanup(t *switchTable) {
-	remove := func(streamID string) {
-		// Helper function to drop a queue
-		buf := b.bufs[streamID]
-		for _, packet := range buf.packets {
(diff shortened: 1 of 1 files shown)


Arceliar:
Tested by lowering the threshold to 256 kB and then checking that traffic could still get through. It seems to work as intended after the last commit.

PR #154, opened 2018-07-07
Add support for UNIX domain admin sockets

neilalexander:
This adds support for UNIX domain sockets for the admin socket to both Yggdrasil and `yggdrasilctl` on supported platforms.

This also adds support for specifying `AdminListen` in URI format, i.e. `tcp://localhost:9001` or `unix:///var/run/yggdrasil.sock`, and adds the same support to the `-endpoint=` switch in `yggdrasilctl`.


neilalexander:
Have tweaked to use `Close` on the listener instead of removing the socket file manually - this seems to work fine on macOS.

Feel free to merge once you've had another look over!

Issue #144, opened 2018-06-22
Multiple paths, disconnected one isn't routed around until after it times out


Arceliar:
Two real nodes, A and B,  are both peered with X and only X. X has a route to the root, so A and B both make X their parent.

Running iperf between A and B, they get something like 7 Mbps. If I connect them directly to eachother over GbE, the speed jumps up to ~850+ Mbps after about 2 seconds. That part works great.

Now I unplug the cable between A and B. The iperf stream's throughput drops to 0 bps until the moment the connection between A and B times out, at which point it goes back to 7 Mbps. What I want to have happen is for the speed to drop down to ~7 Mbps almost instantly, without waiting for the peer to time out. This is a prerequisite for #111 I think.

Node coords shouldn't be changing (they keep the same parent, X, the whole time). What I would have expected is for the stack for link A->B to grow, and the network would switch to sending over A->X->B instead. For some reason, as soon as the write call blocks for the disconnected peer, the node stops trying to send to that peer (the node stops sending to the buffered channel that's used to feed the tcp write worker, which would normally read from the channel to build up the LIFO stack and then pop from the top of the stack when there's no more traffic on the go channel). Pings also fail to make it through, so I don't think it's any issue related to the inner iperf stream backing off or delaying retransmission attempts.

neilalexander:
I am curious whether the backpressure queue changes would have potentially resolved this in part - once a connection starts to block, I would expect that the queue would grow rather quickly, triggering a fairly fast route-around onto another link?

(Once the peer connection has timed out I assume that the queue is discarded rather than drained into another queue? I don't know if we might want to examine that.)

Arceliar:
Regarding your second point, there's no per-peer queue to drain other than whatever the OS manages as part of the TCP buffer, as of that last major set of backpressure changes. Per destination queues are kept in the switch, and are fed do peer connections as those connections become available.

And you're right about the intended behavior, what *should* happen is that a `net.Conn.Write()` call will block, causing it to never inform the switch that it's idle and ready for more traffic. So the switch sould route packets to another interface. What I suspect *really* happens is this:
1. A stream of (TCP) traffic on some path has a link die, lets say it's A->B->C->D and link B->C die.
2. The stream keeps sending, and fills up the buffer (or window? I'm not sure which) for B->C, which causes the last send to block.
3. Because B->C is blocked, the switch directs traffic to follow B->E instead, and we end up with something like A->B->E->D.
[...]

Issue #131, opened 2018-06-15
Testing yggdrasil on macOS 10.13.5 (17F77), ping to local ipv6 fails


satindergrewal:
Using latest darwin binary from here:
https://circleci.com/api/v1.1/project/github/yggdrasil-network/yggdrasil-go/latest/artifacts

binary link: https://156-115685026-gh.circle-artifacts.com/0/yggdrasil-develop-0.2.22-darwin-amd64

- downloaded bin to `~/Downloads` directory
- renamed and changed it's permissions to executable
- generated conf file
- started with sudo command like this:

```shell
$ sudo ./yggdrasil -useconf < yggdrasil.conf 
2018/06/15 20:14:04 Starting up...
2018/06/15 20:14:04 Found 10 multicast interface(s)
2018/06/15 20:14:04 Starting router
2018/06/15 20:14:04 Multicast discovery is enabled
2018/06/15 20:14:04 Listening for TCP on: [::]:36845
2018/06/15 20:14:04 Interface name: utun3
2018/06/15 20:14:04 Interface IPv6: 203:5af4:178a:4da5:dbd:5380:32e3:47ed/7
2018/06/15 20:14:04 Interface MTU: 65535
2018/06/15 20:14:04 Startup complete
2018/06/15 20:14:04 Your IPv6 address is 203:5af4:178a:4da5:dbd:5380:32e3:47ed
2018/06/15 20:14:04 Your IPv6 subnet is 303:5af4:178a:4da5::/64
2018/06/15 20:14:04 Admin socket listening on 127.0.0.1:9001
```

- from other terminal tab, tried to ping to local IPv6 with 100% packet loss

```shell
$ ping6 203:5af4:178a:4da5:dbd:5380:32e3:47ed
PING6(56=40+8+8 bytes) 203:5af4:178a:4da5:dbd:5380:32e3:47ed --> 203:5af4:178a:4da5:dbd:5380:32e3:47ed
^C
--- 203:5af4:178a:4da5:dbd:5380:32e3:47ed ping6 statistics ---
34 packets transmitted, 0 packets received, 100.0% packet loss

$ ping6 203:5af4:178a:4da5:dbd:5380:32e3:47ed
PING6(56=40+8+8 bytes) 203:5af4:178a:4da5:dbd:5380:32e3:47ed --> 203:5af4:178a:4da5:dbd:5380:32e3:47ed
^C
[...]

PR #156, opened 2018-07-07
Version 0.2.4

neilalexander:
I think we're waiting on #144 but this release will include:

- #131 macOS local ping bug fixed
- #154 support for UNIX domain sockets for the admin API
- #155 centralised platform defaults
- #152 limit backpressure resource consumption
- #151 more backpressure tuning

Changelog needs updating before merge.


Arceliar:
Debugging #144 and finding a workaround could take a while, so there's no need to hold off the release if that's the only issue.

I updated the error message from a malformed `ListenAdmin` configuration, so if people run into problems then hopefully the error message is enough to figure out what the problem is.

neilalexander:
I updated for `AdminListen` to fall back to trying TCP, as I was worried that `localhost:9001` may exist in some configs out in the wild, and didn't want to break it. If we want to make a breaking change to force `tcp://` or `unix://` then we could schedule that for v0.3.

Other than that, I'm happy to release if you are.

Arceliar:
Looks good from my side.

PR #170, opened 2018-07-22
Only split queues based on port number for TCP/UDP/SCTP

neilalexander:
Only split queues based on port number for TCP/UDP/SCTP and rely only on protocol number for other protos to prevent issues with IPIP, GRE etc which currently might be split into different queues based on length or other arbitrary values.

We also don't currently accommodate for the fact that the Next Header could be an IPv6 Extension Header, so that might also be desirable for us to check. Currently any IPv6 packet with an EH will be split into queues based on the first EH instead of the payload protocol.


neilalexander:
Incidentally, there are still other problems with this approach - fragments are also identified in the Next Header field (value `44`) amongst other things. With such high MTUs, fragments are not likely, but it may result in them being placed in a different queue to any other traffic from that flow. 

Per-flow classification may actually not be necessary if we already have separate queues by source/destination but it may make sense to pay some closer attention to the Traffic Class or, perhaps even better, the Flow Label fields in the IPv6 header for this purpose, although I suspect that these approaches are not likely to provide an out-of-the-box benefit on an otherwise vanilla system.

PR #176, opened 2018-07-29
Simpler flowlabel parsing; avoid using 0 flowlabel.

cathugger:
some changes which I though would make sense.


cathugger:
Did some tweaks for code and comments.

src/yggdrasil/session.go
```diff
@@ -423,40 +423,37 @@ func (sinfo *sessionInfo) doWorker() {
...
-		// Now we append something to the coords
-		// Specifically, we append a 0, and then arbitrary data
-		// The 0 ensures that the destination node switch forwards to the self peer (router)
-		// The rest is ignored, but it's still part as the coords, so it affects switch queues
-		// This helps separate traffic streams (coords, flowlabel) to be queued independently
-
-		// TODO could we avoid allocations there and put this work into wire_trafficPacket.encode()?
-		coords = append(coords, sinfo.coords...)  // Start with the real coords
-		coords = append(coords, 0)                // Then target the local switchport
+		coords = append(coords, 0)                // First target the local switchport
```
cathugger: is this code multithreaded? if it is, appending to `sinfo.coords` directly could be dangerous, as it could have enough capacity to not cause allocation, and use underlying buffer, and other threads could then overwrite that buffer.
if it is not, then we should ensure that this has enough capacity, to always avoid memory allocation (this is kinda supposed to be hot path I think).

neilalexander: `coords := sinfo.coords` should pass by value and not reference if I am understanding right?

cathugger: when you're working with slices it's by reference

cathugger: well technically it's value copy by slice's value is reference+len+cap tuple

neilalexander: It's not a slice, it's a `[]byte`. Just did a quick sanity check:
```
	if flowkey != 0 {
		coords = append(coords, 0)                // First target the local switchport
		coords = wire_put_uint64(flowkey, coords) // Then variable-length encoded flowkey

		sinfo.core.log.Println("coords: ", coords)
		sinfo.core.log.Println("sinfo: ", sinfo.coords)
	}
```
Result:
```
2018/07/30 13:14:52 coords:  [8 0 174 236 108]
2018/07/30 13:14:52 sinfo:  [8]
```

cathugger: byte slice. it's not byte array. also like I said it depends on current cap of slice.
if available cap is less than stuff you add, reallocation happens.
if it is enough, no reallocation happens.
this case is dangerous in multithreading code, because 2 threads may overwrite same underlying buffer.

Arceliar: The `doSend` function is only called by a single goroutine for any particular session. Another goroutine can update `sinfo.coords` when session ping/pong packets come in, so that part technically isn't thread safe, but those coord changes will update the slice by pointing to another array, so it shouldn't cause multiple append operations to overwrite eachother. I think it *may* be possible for the slice length and the pointer to get out of sync during an update by the other thread, but that probably affects at most 1 packet when it happens. I'm not sure we really need to care. We could always throw a sync.Atomic at it to avoid significantly refactoring the code, and then move ping/pong updates into the per-session worker goroutine at a later date.

PR #188, opened 2018-10-07
Add DSCP classification to TCP

neilalexander:
This PR adds a new `DSCPClassification` configuration option, which allows marking TCP peering traffic for QoS. 

In my testing on macOS so far, this works fine and correctly marks outbound traffic for outbound peerings (with `Dial`) but does not work on inbound peerings at all (with `Accept`) - it seems to produce "`setsockopt: invalid argument`".

It will be interesting to see if this is the same on Linux and other platforms, or if there is something I am missing.

In cases where the local `DSCPClassification` is set to `0`, we may want to try and send traffic back using the same DSCP classification that we accepted the TCP connection with. In cases where the setting has an explicit value, I guess we should respect that value for any traffic we send back.


neilalexander:
Parked until Go has a better way to deal with this - right now it means going to Go 1.11 and using platform-specific syscalls, so it's not guaranteed to work everywhere anyway.

PR #197, opened 2018-10-21
Metadata support

neilalexander:
This adds two features:

- A `FriendlyName` configuration option, which can potentially be sent/received with direct peers and/or sessions
- Friendly names and real endpoints in the admin socket `getPeers`, `getSwitchPeers` etc

I thought about tinkering with `version.go` to exchange the friendly name with peers, but I can't think of a way to do so without breaking connectivity with older peers since they rather explicitly check the expected length on the wire of the version metadata.


neilalexander:
This PR is a mess after trying to merge v0.3 into it. Closing.

PR #201, opened 2018-11-06
Tunnel traffic using crypto-key routing

neilalexander:
This PR allows non-Yggdrasil subnets to be routed over Yggdrasil. It allows for specific subnets to be routed to a specific public key (`IPv4Destinations` and `IPv6Destinations`) and allows specifying valid source addresses for crypto-key-routed traffic (`IPv4Sources` and `IPv6Sources`).

### DIFF
--- a/src/yggdrasil/router.go
+++ b/src/yggdrasil/router.go
@@ -259,6 +259,12 @@ func (r *router) recvPacket(bs []byte, sinfo *sessionInfo) {
 		util_putBytes(bs)
 		return
 	}
+	var dest address
+	copy(dest[:], bs[24:])
+	if !r.cryptokey.isValidSource(dest) {
+		util_putBytes(bs)
+		return
+	}
 	var source address
 	copy(source[:], bs[8:])
 	var snet subnet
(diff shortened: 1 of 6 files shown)


neilalexander:
I've had the opportunity to go over this again today and to perform some more extensive testing between containers, and a number of bugs have been fixed.

Following a review I think this is good to be merged - the caveat is that it's IPv6-only for now, and we'll have to worry about IPv4 later. I also think we can improve the configuration in the future as well as a part of v0.3.

src/yggdrasil/router.go
```diff
@@ -121,19 +125,39 @@ func (r *router) sendPacket(bs []byte) {
...
+			copy(snet[:], addr[:])
+			if !dest.isValid() && !snet.isValid() {
+				return
+			}
+		} else {
+			return
+		}
+	}
 	doSearch := func(packet []byte) {
 		var nodeID, mask *NodeID
```
Arceliar: When creating a session for CKR routes, we should use the full key, not just the part we can see from the IP address. That would mean setting the nodeID and mask outside of `doSearch`. Similarly, for CKR routes, when we look up a session, we should check that the session matches the CKR key, not just the IP address.

neilalexander: I've addressed this in fbfae47.

PR #203, opened 2018-11-10
Chord

Arceliar:
This patch switches from a kad-based DHT to a chord-based one. This is *technically* backwards incompatible, in the sense that a network with mixed DHT implementations isn't guaranteed to converge, but the wire formats don't change, and searches still seem to succeed in a network with the old DHT nodes (though they seem to take a little longer, I think due to getting unexpected results).

The main advantage of the chord-based approach is that only O(1) state is required for correctness, the rest is just to make searches finish faster (or peers, for bootstrapping reasons, but we know about them for free). This opens the door for future optimizations, such as reducing idle DHT traffic, or prioritizing successor and predecessor to make nodes reachable more quickly after a coord change.

### DIFF
--- a/misc/sim/treesim.go
+++ b/misc/sim/treesim.go
@@ -415,10 +415,10 @@ func main() {
 	}
 	fmt.Println("Test")
 	Util_testAddrIDMask()
-	idxstore := makeStoreSquareGrid(4)
+	//idxstore := makeStoreSquareGrid(4)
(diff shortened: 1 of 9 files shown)


Arceliar:
Marking unfinished just so we don't merge it right away. I'm thinking we give the current develop a week or so to test, to make sure everything works without any issues with the new DHT getting in the way.

Issue #173, opened 2018-07-25
Latency optimisation


neilalexander:
Currently there is nothing to stop some nodes joining the network, peering with nodes that have unacceptable latency and significantly impacting the latency of traffic flowing through these routes in the process. This issue is largely just to track this problem, and to keep a note of some possible potential fixes for the future.

Some possible ideas:

#### 1. Including latency in the coordinate selection metrics as well as link uptime/stability
At the moment, when node coordinate selection takes place, the link uptime/stability is the primary metric involved. This keeps the topology of the network relatively stable even if not all peerings are solid. However, there is nothing to account for the actual latency of each peering, so there is no guarantee that your chosen coordinates will be aligned to your physically closest node.

In the event that multiple peerings are considered "stable", we should aim to align node coordinates to the node with the lowest latency. This at least will encourage the worst-path route to follow the path of shortest latency.  Some work will need to be done to identify how much weight wants to be given to latency vs stability if the link is not 100% stable.

Also cautious that we don't generate too much extra traffic as a result (#65).

Currently blocked by:
- There is no ping traffic between peers that can be used to calculate the latency

#### 2. Share latency information through backpressure notifications
It might be possible to share latency information through backpressure notifications, if that happens as a part of #111. 

Currently blocked by:
- There is no ping traffic between peers that can be used to calculate the latency
- There is no backpressure notification traffic either

#### 3. Add a `LatencyThreshold` option for peerings
A non-default user-configurable option, in ms, to define the maximum allowed latency on an incoming peering. This might be particularly useful for public nodes to prevent peerings being accepted from nodes that are far away. The ping RTT time must fall beneath the `LatencyThreshold` value or the peering will be rejected/dropped by the TCP machinery before the peering negotiates successfully.

Currently blocked by:
- There is no ping traffic between peers that can be used to calculate the latency

Arceliar:
If we don't want to explicitly measure latency, we can implicitly infer which path has the lowest latency to the root based on who we receive coord updates from first whenever the root updates its timestamp. Sort-of like how B.A.T.M.A.N. uses whichever route delivers the OGM messages first. The thing is, we don't want coords to flap needlessly, so we'd need to keep some kind of rolling average and then include it as *part* of the criteria we use to decide on a parent node (along with stability, and either with or replacing path length).

Casperpor:
[...]

PR #208, opened 2018-11-17
Latency optimization for switch coord selection

Arceliar:
The basic idea here is to do two things:
1. When the root updates a timestamp, keep an exponentially weighted rolling average of how much time passes between receiving the updated timestamp from each peer, relative to the time the current parent sends it.
2. When the parent disconnects or changes coords, switch your new parent to whichever node has the lowest average latency.

There should probably be some point at which another node is stable enough and low latency enough that we should set them as the parent immediately, instead of waiting for the current parent's state to change. It needs to be more than simply having lower latency, because that seems to cause too much route flapping when there are multiple peers with similar latency (particularly for large-ish networks in the sim, where latency is smaller than the variation from the runtime's goroutine scheduler).

A complete version of this (with the remaining issue fixed) would count as a fix for #173 I think.

### DIFF
--- a/src/yggdrasil/switch.go
+++ b/src/yggdrasil/switch.go
@@ -118,13 +118,13 @@ func (x *switchLocator) isAncestorOf(y *switchLocator) bool {
 
 // Information about a peer, used by the switch to build the tree and eventually make routing decisions.
 type peerInfo struct {
-	key       sigPubKey     // ID of this peer
-	locator   switchLocator // Should be able to respond with signatures upon request
-	degree    uint64        // Self-reported degree
-	time      time.Time     // Time this node was last seen
-	firstSeen time.Time
-	port      switchPort // Interface number of this peer
-	msg       switchMsg  // The wire switchMsg used
+	key     sigPubKey     // ID of this peer
+	locator switchLocator // Should be able to respond with signatures upon request
+	degree  uint64        // Self-reported degree
+	time    time.Time     // Time this node was last seen
+	cost    time.Duration // Exponentially weighted average latency relative to the current parent, initialized to 1 hour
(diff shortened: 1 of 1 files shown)


Arceliar:
Qualitatively speaking, I think this now does the right thing.

Instead of keeping a rolling average, latency relative to the current parent is tracked cumulatively for each peer. For example, if a switch message with a new root timestamp arrives 10ms after the announcement from the current parent, then 10ms is added to that peer's counter. If a peer is faster than the current parent, then this number will be negative. The total is bounded to +- 1 second.

If the cumulative latency ever reaches the minimum (-1 second), then we switch to using that peer as a parent. To give some numbers, assuming a new peer comes online and the counter is initialized to 0, then if the peer is on average 1ms faster than the current parent, this will take about 8.5 hours to switch. In the same scenario, if it's 50ms faster, then this will take about 10 minutes. The maximum a single message can change the total is +- 1/8 second, which takes about 4 minutes to switch. The relatively slow / cumulative nature means a peer needs to be consistently faster than the parent for a while before switching, and the maximum change is bounded, so this should prevent things from oscillating constantly when two peers have similar latency. The cumulative total gets reset to 0 if a peer's coords change, so it will still tend to select stable peers first.

The 1 second threshold, and the maximum 1/8th second change per message, are things that could be tuned further in the future. In particular, I imagine it could lead to flapping for node that exclusively have high latency paths to the root, such as a tor-only node with many peers.

Arceliar:
Closing this PR since latency is so stateful. We should continue to think about sane ways to handle this, but the current shortest-path approach isn't so bad as to limit usability, and, in the absence of a strong theoretical justification for the approach we use, I'd rather have something simple and easy to understand.

PR #210, opened 2018-11-21
Admin functions for cryptokey routing

Arceliar:
Work in progress. This moves the `router.recvPacket` calls into the main router goroutine, so we can keep all of the `router.cryptokey` code running in that one goroutine. It then adds admin API calls for: `addSourceSubnet`, `addRoute`, `getSourceSubnets` and `getRoutes`, though `yggdrasilctl.go` doesn't have pretty printing for the responses to the "get" functions.

TODO: `removeSourceSubnet` and `removeRoute`, and any little things like more descriptive API function/argument names.

The goal is that users should be able to check/add/remove cryptokey routes and source subnets at run time. Then people could hypothetically write some 3rd party application to e.g. let users register to get a public IP/subnet, and add those subnets and routes, without needing to restart every time something is added or removed.

### DIFF
--- a/src/yggdrasil/admin.go
+++ b/src/yggdrasil/admin.go
@@ -456,7 +456,6 @@ func (n *admin_nodeInfo) toString() string {
 		out = append(out, fmt.Sprintf("%v: %v", p.key, p.val))
 	}
 	return strings.Join(out, ", ")
-	return fmt.Sprint(*n)
 }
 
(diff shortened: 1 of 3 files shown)


Arceliar:
Added `removeSourceSubnet` and `removeRoute` functions, and some formatting in `yggdrasilctl` for `getSourceSubnets` and `getRoutes`, so I think this covers most of the important stuff now.

PR #212, opened 2018-11-25
Add dhtPing to the admin interface

Arceliar:
This refactors the DHT code a bit so, instead of passing all DHT responses to the search code, other parts of the code (including searches) can register a callback for a particular key/destination pair when they send a DHT lookup.

Then rest of this PR adds a `dhtPing` call to the admin interface, which uses the above callback mechanism to wait for a response from a remote node, or else returns an error after a timeout. The `dhtPing` function takes a (hex-encoded) target NodeID as an optional argument. If not specified, the target NodeID defaults to the ID of the remote node being asked to do the lookup (which returns nothing for kad nodes, but should be sufficient to walk the whole network with the new chord dht).

Some details about how the timeout works, and the response format, could maybe use some work. The code otherwise seems to work as expected.

### DIFF
--- a/src/yggdrasil/dht.go
+++ b/src/yggdrasil/dht.go
@@ -361,6 +361,9 @@ func (t *dht) getImportant() []*dhtInfo {
 
 // Returns true if this is a node we need to keep track of for the DHT to work.
 func (t *dht) isImportant(ninfo *dhtInfo) bool {
+	if ninfo.key == t.core.boxPub {
+		return false
+	}
 	important := t.getImportant()
 	// Check if ninfo is of equal or greater importance to what we already know
 	loc := t.core.switchTable.getLocator()


src/yggdrasil/admin.go
```diff
@@ -536,6 +556,7 @@ func (a *admin) getData_getSelf() *admin_nodeInfo {
 	table := a.core.switchTable.table.Load().(lookupTable)
 	coords := table.self.getCoords()
 	self := admin_nodeInfo{
+		{"key", hex.EncodeToString(a.core.boxPub[:])},
```
neilalexander: In other places in the admin socket, we use `box_pub_key` instead of `key` as the key name. I am not sure if we want to continue using that convention for clarity?

PR #214, opened 2018-12-01
Give a lower priority to (outgoing) socks connections.

Arceliar:
The basic idea here is that peers get a priority. If peer A's priority is higher than peer B, then we'll always prefer peer A as a parent instead of peer B, as long as it meets the other requirements (no loops, etc).

Priority defaults to 0. Currently, we set priority to 1 if we see that it's *not* a `wrappedConn`, so direct connections should get priority over `socks://`. Something should also be done to check if an incoming connection is proxied, but I'm not sure what.

If we decide to go this route, we can consider giving users some way to manually set priority for specific peers / kinds of peers, but this is one of those things where doing it wrong can hurt other people's network performance, so we probably need to be careful about that.

### DIFF
--- a/src/yggdrasil/peer.go
+++ b/src/yggdrasil/peer.go
@@ -88,6 +88,7 @@ type peer struct {
 	endpoint     string
 	friendlyName string
 	firstSeen    time.Time       // To track uptime for getPeers
+	priority     uint8           // Higher priority -> prefer as a parent in the tree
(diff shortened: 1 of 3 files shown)


Arceliar:
This is basically an alternative to #208, except it only covers a particularly bad scenario, and only under special cases. I'm going to look into latency-based parent selection more, to see if I can come up with something that reliably picks the fastest parent (or at least avoids particularly slow parents) without causing too much route flapping.

PR #216, opened 2018-12-02
Add SwitchOptions and MaxTotalQueueSize

neilalexander:
This adds the ability to control the maximum size of all backpressure queues. Currently it defaults to 4 megabytes.

We may also want to be able to cap the maximum allowable number of queues, or cap the maximum size of each individual queue too in the future.


Arceliar:
I think capping individual queue size shouldn't matter, since packets get (probabilistically) dropped from longer queues when things are congested. Adding a limit may even encourage you to use multiple flow IDs, to try and bypass the per-queue limit, and I'd rather not give people any motivation to game the system. Same for limiting number of queues: whatever you do to limit it, you're potentially allowing DoS, either by people flooding random flows to push out existing queues, if old queues get dropped to make room, or flooding random flows to take up any empty slots if new queues are dropped.

PR #218, opened 2018-12-03
Docker Support & Updated License

cwchristerw:
### DOCKER SUPPORT
INFO :information_source: 
These files shouldn't need much maintaining because it uses directly newest version. I would suggest that in future you'll create yggdrasil-docker repository for supporting creating yggdrasil with different releases (when first one is out) and develop. And also add to https://hub.docker.com/

YOUR KEY TO SUCCESS TESTING
Build image from source code
`sudo docker build -t yggdrasil-network/yggdrasil-go -f $(pwd)/contrib/docker/Dockerfile $(pwd)`

Create container from image
`sudo docker run --name yggdrasil -d --net=host --cap-add=NET_ADMIN --device=/dev/net/tun -v $(pwd)/contrib/docker/config:/etc/yggdrasil-network --restart always yggdrasil-network/yggdrasil-go`

DOCUMENTATION :paperclip: 
I can help making better documentation but these are the commands that you will need to test in your machine.

THANKS :smile: 
Special thanks to Lars Gierth <larsg@systemli.org> (@lgierth) for making Dockerfile for cjdns. I used it as a guide to creating Docker support for Yggdrasil.

### UPDATED LICENCE
Accidentally changed file name and changed back to original. Updated URL from http://fsf.org/ to https://fsf.org/


hwpplayer1:
I do not force people for licensing but I am keen to learn licensing so how about LGPLv3+ option for that project ?

PR #221, opened 2018-12-04
Reduce container image size

jcgruenhage:


jcgruenhage:
Inside a running container we only have access to what is available in it, and since this has a startup script we at least need a shell. For debugging purposes it's also helpful if there is a remotely usable system available in the container. Therefore, you usually put a full OS - Kernel in there, in this case it's alpinelinux, which is ~5MB.

If we get rid of the startup script and let the user generate the config manually, we could have a container image that has nothing but the yggdrasil binary in it.

cwchristerw:
I would suggest we don't delete startup script because it would be nice to have almost ready to play instance

jcgruenhage:
Yeah, I don't want to get rid of it, just wanted to mention what is left in the image and for what.

PR #223, opened 2018-12-05
Try to SO_REUSEPORT on multicast socket

### LINKED ISSUE #222: Multicast cannot be enabled on more than one Yggdrasil process (port reuse)
When running multiple instances in same machine, admin port should not be static. It would be nice increase by one when port is reserved instead of panic :smile:. Then should automatically change new free port to configuration. Could be temporarily fixed by editing config manually when first started.

2018/12/05 21:00:52 Starting up...
2018/12/05 21:00:52 Found 5 multicast interface(s)
2018/12/05 21:00:52 Starting switch
2018/12/05 21:00:52 Starting router
2018/12/05 21:00:52 Multicast discovery is enabled
2018/12/05 21:00:52 Failed to start multicast interface
2018/12/05 21:00:52 An error occurred during startup
panic: listen udp6 [::]:9001: bind: address already in use

goroutine 1 [running]:
main.main()
	/src/yggdrasil.go:227 +0x15bc

neilalexander:
This PR aims to fix #222 by setting `SO_REUSEPORT` (or `SO_REUSEADDR` on Windows) on the multicast socket. This allows multiple Yggdrasil process to run side-by-side with multicast support enabled, as long as the `Listen` and `AdminListen` directives are configured differently on each instance.

Currently this only adds support for Linux, Darwin, Windows and the BSDs. On other platforms (Solaris, Plan9) nothing is done and reusing the multicast socket will fail in the usual way.

This PR does use `ListenConfig` which I believe is only present in Go 1.11, so we may be tying ourselves to requiring Go 1.11 to build Yggdrasil. I don't suppose this is a problem, and we may already do that in other places, but worth mentioning regardless.

### DIFF
--- a/src/yggdrasil/multicast.go
+++ b/src/yggdrasil/multicast.go
@@ -1,6 +1,7 @@
 package yggdrasil
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"time"
@@ -35,7 +36,10 @@ func (m *multicast) start() error {
 			return err
 		}
 		listenString := fmt.Sprintf("[::]:%v", addr.Port)
-		conn, err := net.ListenPacket("udp6", listenString)
+		lc := net.ListenConfig{
+			Control: multicastReuse,
+		}
+		conn, err := lc.ListenPacket(context.Background(), "udp6", listenString)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


cwchristerw:
Does this make sure that both instances can talk to each other or does the second just replace first one.

neilalexander:
@cwchristerw If `MulticastInterfaces` are set on both instances then they will automatically peer with each other.

cwchristerw:
@neilalexander That great :smile:

PR #227, opened 2018-12-08
convert to go module

Arceliar:
Pretty much what the title says.

Files under `src/yggdrasil/config` were moved to `src/config` and likewise for `defaults`.

The code for the executable files can now be found in `cmd/yggdrasil/main.go` and `cmd/yggdrasilctl/main.go`, and the build script has been updated accordingly.

Basically, `go build -tags debug ./...` doesn't produce errors, and `./build` gives us binaries in the usual spot, so this doesn't have really obvious problems that I can see, but I don't know what else needs testing. There's probably still at least some things that don't work right but could.


Arceliar:
Apparently the CI doesn't like something. Not really a surprise.

PR #244, opened 2018-12-12
Compile really static with no pie

jsarenik:
On Alpine Linux (musl-libc based GNU/Linux distribution)

Before: $ file yggdrasil
yggdrasil: ELF 64-bit LSB pie executable x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, stripped

After: $ file yggdrasil
yggdrasil: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped

### DIFF
--- a/build
+++ b/build
@@ -5,6 +5,7 @@ PKGNAME=${PKGNAME:-$(sh contrib/semver/name.sh)}
 PKGVER=${PKGVER:-$(sh contrib/semver/version.sh --bare)}
 
 LDFLAGS="-X $PKGSRC.buildName=$PKGNAME -X $PKGSRC.buildVersion=$PKGVER"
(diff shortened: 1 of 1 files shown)


jsarenik:
From IRC:

> So for Alpine you should be able to do `./build -l "-static -no-pie"` to build Ygg

jsarenik:
Actually this works on `master` or `develop` branch:

```
./build -l "-extldflags '-static -no-pie'"
```

jsarenik:
On current master (0.4.0) one has to do following:

```
./build -l "-linkmode external -extldflags '-static -no-pie'"
```

See https://github.com/jsarenik/yggdrasil-go/tree/jasan/static

PR #246, opened 2018-12-12
NodeInfo support

neilalexander:
To-do:

- [x] A way to send metadata through the admin API, probably with some callback so that you can get the result back
- [x] Caching of metadata in somewhere sensible - perhaps a dedicated metadata cache, which maps `box_pub_key`s to `metadata` so that you can go back and see the metadata of peers, sessions etc
- [x] Using `Metadata` specified in the configuration file (in flexible map format)
- [x] A sensible format to agree on:
  - Probably JSON makes sense, since we can already put HJSON/JSON into the configuration file and easily transform it into real JSON if not already
  - Come up with some standardised stanzas in that case
    - For service discovery we could have something akin to SRV records (e.g. a `name`, `protocol`, `portnumber` tuple) so that crawlers can find services and work out what they are without port-scanning
    - example: `{ "services": [ {"name": "http", "proto": "tcp", "port": 80}, {"name": "irc", "proto": "tcp", 6667}, {"name": "dns", proto: "udp", "port": 53} ] }`
  - Upper size limit for metadata
- [x] Perhaps a way to exchange metadata using permanent public keys without triggering a full session being opened (although this might take some refactoring)


Arceliar:
> - [ ] Perhaps a way to exchange metadata using permanent public keys without triggering a full session being opened (although this might take some refactoring)

I strongly recommend we make this a top priority, before the other to-do items, since there's no a priori reason why we should need to set up a session if you're not sending traffic *through* the session. Best to get the refactoring out of the way up front. I thing this makes sense too, the request can look a lot like the current `dhtPing` approach, except ygg (probably) isn't using the same packets in the background on its own. I haven't looked a the PR *too* closely yet, but I still have the `dhtPing` stuff pretty fresh in my memory, so I think we can probably refactor this without too much trouble.

PR #251, opened 2018-12-15
Tune peering

Arceliar:
Does two things right now:
1. Immediately send a switch message when we first connect to a peer.
2. Tell the OS to send TCP keepalive traffic. We currently send application-level keepalive traffic of our own, at least as often as the threshold we give the OS, so this shouldn't actually do anything yet. Long-term, I'd like to switch to having the OS be responsible for this, except in cases where it can't be, since this will let us produce less idle keepalive traffic per peer.


neilalexander:
Works great between two nodes running this branch - they converge pretty much instantly.

PR #270, opened 2018-12-19
Don't fail if tag exists

neilalexander:
This will stop the CircleCI pipeline failing if the tag already exists.


neilalexander:
Minor tweak - merge when ready.

PR #275, opened 2018-12-20
Allow hiding nodeinfo defaults

neilalexander:
This creates a keyword `null` for users who specifically want to hide built-in nodeinfo defaults like `buildplatform` and `buildarch`, for example:
```
NodeInfo:
{
  buildplatform: null
  buildarch: null
}
```
One or two individuals have suggested to me that they would prefer to be able to hide those as they could be personally identifying in some ways. 

Additionally, I don't particularly want to make it easy to *skew* those values by entering false ones, which is why I haven't added those built-ins to the generated default configuration. It'd be better to lack the data than have wrong data.


src/yggdrasil/nodeinfo.go
```diff
@@ -108,9 +108,14 @@ func (m *nodeinfo) setNodeInfo(given interface{}) error {
 	if nodeinfomap, ok := given.(map[string]interface{}); ok {
 		for key, value := range nodeinfomap {
 			if _, ok := newnodeinfo[key]; ok {
+				if value == "hide" {
```
cathugger: should `strings.EqualFold`, to allow any case of "Hide" "HIDE" etc.
also would this even compile, as value will probably be of type `interface{}` not `string`?

src/yggdrasil/nodeinfo.go
```diff
@@ -108,9 +108,14 @@ func (m *nodeinfo) setNodeInfo(given interface{}) error {
...
+					delete(newnodeinfo, key)
+				}
 				continue
 			}
-			newnodeinfo[key] = value
+			if value != "hide" {
```
cathugger: maybe this could be moved up and folded into one check overally, something like `if s, ok := value.(string); ok && strings.EqualFold(s, "hide") { delete(newnodeinfo, key); continue }`

Arceliar:
Since there's no way to guarantee that the information is accurate (you could always compile your own client that lies), maybe we shouldn't force this info to be there in the first place. We could just add it to the defaults for nodeinfo or something...

src/yggdrasil/nodeinfo.go
```diff
@@ -108,9 +108,14 @@ func (m *nodeinfo) setNodeInfo(given interface{}) error {
 	if nodeinfomap, ok := given.(map[string]interface{}); ok {
```
cathugger: unrelated, but if `given` is always expected to be of type `map[string]interface{}`, maybe it should be just cast without ok check

cathugger:
>We could just add it to the defaults for nodeinfo or something...
that wouldn't work with client versions or build changes if using same config

cathugger:
this sorta would disallow using word "hide" in any of metadata values, maybe we should just utilize something like json null or false for this? imo null would be pretty okay for blacklist

cathugger:
unless values are always forced to be string, then kinda no way around it

Arceliar:
Maybe we should add a second NodeInfo-related option to the config. Just a bool flag. If it's true (the default value), we add the automatic info to the NodeInfo, and if it's false, then we don't (we *maybe* delete any fields with the same name from the NodeInfo, just to avoid confusion / make it not completely trivial to lie, so we can still rely on platform and version info to be accurate on average in the cases where it's there).

cathugger:
This mainly came up as anonymity sensitive thing, especially regarding buildarch and buildplatform, but buildname and buildversion values aren't that anonimity-sensitive, and excluding them may make some debugging harder imo, so im not really sure about flag to exclude everything

cathugger:
but I guess for ones who really want to block everything, keeping track of list of options ygg by default includes would be more annoying than setting one flag.
so yeah maybe that'd be right approach, combined with null for explicitly deleting individual keys.

Arceliar:
I think I also have a preference for `null` as a way to intentionally disable one item in the NodeInfo, since it seems more natural than a magic value like `"hide"`.

neilalexander:
I've added the `NodeInfoPrivacy` boolean option for hiding the built-in defaults, and it now allows `null` or `"null"` for selectively hiding the defaults if not wanting to hide all of them.

src/yggdrasil/nodeinfo.go
neilalexander: It can be `null`.

PR #276, opened 2018-12-21
Tune DHT a little better

### LINKED ISSUE #271: "Last seen" DHT info for peers suggests something's buggy
The "last seen" field from `yggdrasilctl` prints when a DHT node last responded to one of our pings. The one ugly bit is for direct peers, where we need to insert them into the DHT somehow to get started. This is supposed to keep the relevant field(s) intact, so these nodes will get pinged periodically (and the pings will throttle down over time), but the counter seems to always be 0-1s when I monitor things with `yggdrasilctl`.

I suspect that either the time isn't getting copied right, or something in the delay isn't getting increased correctly for peers, so things keep resetting back to 0 (and may be pointlessly pinging peers every second or so, I haven't had a chance to investigate yet).

### LINKED ISSUE #273: Reduce idle DHT traffic
This is ignoring any extra traffic from bugs, such as possibly #271.

Currently, once per second, we iterate over the nodes in the DHT and decide which ones (if any) to ping. Specifically, we only ping "important" ones (some function of key space and relative distance on the tree that's expected to tend towards O(logn) size and lookup time, but lead to lower latency lookups on average than the "normal" chord finger table), and the interval between pings increases by 1 second with every ping, up to a maximum of 30 seconds.

Instead, what I'd like to do is add a "dirty" field to each DHT entry. A new node in the DHT starts as clean, and the maximum time before pinging a clean node would keep increasing to something like ~5 minutes. If some other node asks us to do a lookup (for a node ID other than that node's own ID, i.e. it's not just pinging us), then we should mark all the nodes in the response as dirty. Dirty nodes should get pinged if we haven't heard from them in something more like the current 30 seconds, to make sure that we check on nodes that we've been telling other people about (to avoid sending useless info that could lead to pointless checks and timeouts on their end), and significantly reduce idle bandwidth consumption when nobody's asking. Exact numbers for thresholds may be different, I'm just trying to ballpark it. We should also switch from a linear delay increase to an exponential one, so ping traffic backs off faster.

EDIT: part of #65

Arceliar:
Does two things:
1. Moves the peer-specific logic for adding nodes to the DHT from router.go to dht.go and fixes #271 by adjusting how peers get inserted into the DHT (they're only really needed for bootstrapping reasons). I'm not sure why the old version wasn't working right, but I think it was spamming peers with DHT traffic more often than needed, so this may also happen to reduce idle bandwidth.
2. Tries to fix #273 by adjusting how DHT throttle works. Things will now throttle back exponentially. If the DHT is in use, it should ping nodes after at most 30 seconds, which is how things currently work in the master branch. However, if DHT info about a node has *not* been used (specifically, if you haven't sent it as a response to a node that asked about an ID other than their own), then maximum the throttle backs off has been increased to 5 minutes, with timeouts after 6 minutes. This should decrease idle bandwidth use (from the DHT) by a factor of ~10. However, this needs testing, as it currently still seems to ping all nodes after 30 seconds (I suspect this is from idle traffic sent by old kad nodes, which would flag dht info as being used when it's really not).

I need to add some debug output and run some more tests, to try and confirm that old kad nodes are responsible for part 2 not working right, but I'm opening the PR now just to document what's going on / let someone else try testing if I don't manage to get to it in the next few days.

### DIFF
--- a/src/yggdrasil/dht.go
+++ b/src/yggdrasil/dht.go
@@ -134,6 +134,15 @@ func (t *dht) insert(info *dhtInfo) {
 	t.table[*info.getNodeID()] = info
 }
 
+// Insert a peer into the table if it hasn't been pinged lately, to keep peers from dropping
+func (t *dht) insertPeer(info *dhtInfo) {
+	oldInfo, isIn := t.table[*info.getNodeID()]
+	if !isIn || time.Since(oldInfo.recv) > 45*time.Second {
+		// TODO? also check coords?
+		t.insert(info)
+	}
+}
+
 // Return true if first/second/third are (partially) ordered correctly.
 func dht_ordered(first, second, third *crypto.NodeID) bool {
 	lessOrEqual := func(first, second *crypto.NodeID) bool {
--- a/src/yggdrasil/dht.go
+++ b/src/yggdrasil/dht.go
@@ -325,7 +325,15 @@ func (t *dht) doMaintenance() {
 	}
 	t.callbacks = newCallbacks
 	for infoID, info := range t.table {
-		if now.Sub(info.recv) > dht_timeout || info.pings > 6 {
+		switch {
+		case info.pings > 6:
+			// It failed to respond to too many pings
+			fallthrough
+		case now.Sub(info.recv) > dht_timeout:
+			// It's too old
+			fallthrough
+		case info.dirty && now.Sub(info.recv) > dht_max_delay_dirty && !t.isImportant(info):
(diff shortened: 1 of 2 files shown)


Arceliar:
OK, didn't need to do anything, the (I assume old/kad) nodes that were triggering the refreshes seemed to have mostly stopped, so now there's only 1 old kad node causing stuff to get marked about once per minute. I honestly can't think of a better test. Seems to reduce bandwidth usage quite a bit too.

PR #287, opened 2018-12-26
Add flexible logging support

neilalexander:
This uses the `github.com/gologme/log` package for flexible logging support. It gives us the ability to use "trace", "debug", "info", "warn" and "error" levels in addition to the current `Println`/`Printf` logging we have now. 

I think we possibly only need "debug" and "trace", since we already don't make a lot of other input that's interesting.


neilalexander:
This does create some merge conflicts with the `link` branch, but I suggest we merge this first and then `link` will be easy enough to fix after.

PR #301, opened 2019-01-02
Boilerplate for iOS support

neilalexander:
This PR includes some boilerplate code that is needed to support iOS targets.

`Gomobile` is very particular about types - it won't generate bindings in `Yggdrasil.framework` for anything that uses a non-standard type, or even channels for that matter.


neilalexander:
I think this is probably fine for now - most of the work I can think of that I need to do is on the Swift side.

PR #303, opened 2019-01-09
Spelling fixes (Round 2)

deavmi:
Second round of spelling fixes because this time I will make better pull requests.

This doesn't invalidate round 1.


deavmi:
Do not use this pr as I was working using `master` and not develop.

PR #306, opened 2019-01-13
AWDL support, Android tweaks

neilalexander:
This makes some minor changes in the name of mobile support, and also amazingly gets AWDL support to work on macOS. Hopefully it'll work for iOS too.


neilalexander:
I think this should be alright for now - it sort of gets AWDL working on macOS (even if not on iOS yet), fixes multicasting so that it uses the address zone for source interface (which I think is correct behavior) and adjusts the AWDL functions so they perform a regular TCP-like exchange rather than expecting the keys to be known ahead of time.

PR #318, opened 2019-01-27
Implement link interface

neilalexander:
This PR implements `link.go`, from which `tcp.go` and `awdl.go` are now using.

There are not really any user-facing changes with this but it should make it easier to implement new transports in the future.


neilalexander:
To-do list:

- [ ] Checking `intf.link.core.config.AllowedPublicEncryptionKeys` when we receive a non-local incoming connection [like here](https://github.com/yggdrasil-network/yggdrasil-go/blob/b4a7dab34da3a746518916b710e1f4842840e486/src/yggdrasil/tcp.go#L267-L275) (protected by `intf.link.core.configMutex`)

PR #328, opened 2019-02-09
Switch improvements(?)

Arceliar:
Reworks the switch logic.

Previously, if there was more than one node equally close to the destination, then it would break ties arbitrarily by preferring to forward to whichever node it looked at first (random, due to random go map iteration order).

Now, if two nodes are equally close, then prefer to the one with shorter `coords`, as this should tend to favor approaching from the parent side of a node rather than via one of their children. Since nodes select their parents, but not their children, this should (hopefully) give more consistent behavior. If two nodes are *also* have equal length `coords`, then break ties consistently based on whichever peer has the lowest port number.

Technically, we could forward to *any* node that's closer to the destination than we are, not necessarily the one that's closest by the tree metric, so there's some flexibility in how we do this. The approach here should avoid breaking anything that isn't always broken, I think, but it would still forward via a descendant if we know a shortcut that results in fewer total hops at the switch level. I'm not sure what the correct behavior is in that case (we don't want to accidentally trick it into always routing by the tree, otherwise a lot of things will break down).

Still to do: test on nodes with many peers, see if it makes sense to relax the forwarding rules any more than we already have.

### DIFF
--- a/misc/sim/treesim.go
+++ b/misc/sim/treesim.go
@@ -6,13 +6,15 @@ import "os"
 import "strings"
 import "strconv"
 import "time"
-import "log"
 
 import "runtime"
 import "runtime/pprof"
 import "flag"
 
+import "github.com/gologme/log"
+
 import . "github.com/yggdrasil-network/yggdrasil-go/src/yggdrasil"
+import . "github.com/yggdrasil-network/yggdrasil-go/src/crypto"
 
 ////////////////////////////////////////////////////////////////////////////////
 
--- a/src/yggdrasil/switch.go
+++ b/src/yggdrasil/switch.go
@@ -645,20 +645,39 @@ func (t *switchTable) handleIn(packet []byte, idle map[switchPort]struct{}) bool
 		ports[0].sendPacket(packet)
 		return true
 	}
+	table := t.getTable()
 	var best *peer
 	var bestDist int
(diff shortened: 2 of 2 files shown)


neilalexander:
I expect this will largely undo the load balancing effect that we currently have when multiple peerings are active across different interfaces to the same node?

Arceliar:
The load balancing *should* still work, because this applies only to idle links. But what *would* probably change is that there would be a consistent order among the links. So if A and B are both idle, A would always get selected first, and then B. The old behavior is to break this tie randomly on a per-packet basis. I haven't tested this yet, to see exactly what the effects are, but it seems like the node should still be able to forward traffic across all links. If anything, this may lead to *higher* bandwidth, since packets are more likely to follow a single path (and avoid reordering) in cases where thing are *not* too congested. Otherwise, it's possible that a consistent order (somehow) causes the user's traffic to throttle down earlier, but I'm not seeing an obvious reason why TCP would like this *less* than the current random setup. But that's one of the main things that needs testing.

One other thing, I think it probably leads to more stable latency in uncongested scenarios. Not necessarily *lower* latency, but less jitter, since the same path would now be used for every packet, unless congestion demands otherwise.

Arceliar:
Superseded by #338 for now.

PR #332, opened 2019-02-10
Improve yggdrasilconf in contrib

neilalexander:
This PR aims to be an overhaul of the `yggdrasilconf` utility in `contrib` so that we can flexibly read, write and manipulate configuration in either HJSON or JSON format, whilst preserving correct formatting and types. This is used by the `vyatta-yggdrasil` package and may be a good candidate for other projects too.


neilalexander:
Moving this to v0.3.4 as I'm probably going to scrap it and start again and I don't think this should block the release of v0.3.3.

PR #333, opened 2019-02-11
Simplifying Dockerfile

cwchristerw:
### DOCKER SUPPORT
With this change you only have to copy this Dockerfile as is and it will use git to clone all files needed.


neilalexander:
Does this perhaps limit to cloning from our own repository, rather than forks?

cwchristerw:
This will clone from this repository rather than fork but it's possible to create separate repository for these changes and leave old almost untouched but there was some file permission issues and that caused error if entrypoint.sh, build and some other file (I dont remember) was not 555.

cwchristerw:
In old one you would have to manually download or clone github repository, and then run commands from #218, but with this it's in image

neilalexander:
I don't really know anything about Docker so if @Arceliar is happy then I'm happy.

cwchristerw:
Also this from #218 with little change.

I would suggest you'll create docker repository for Docker Image stuff because someone may want to build older versions from releases or development.

And that is not possible with these newer changes because instead of using local copies this will always clone newest github repository.

Also there is no way to get development version because it uses automatically master. Using separate repository for docker makes easier to upgrade thing more faster in times when it's needed.

contrib/docker/Dockerfile
```diff
@@ -1,18 +1,19 @@
-FROM docker.io/golang:alpine as builder
+FROM golang:alpine as builder
```
jcgruenhage: Removing the registry here doesn't simplify the Dockerfile, it creates ambiguity, since different distributions and tools might default to different registries.

contrib/docker/Dockerfile
```diff
@@ -1,18 +1,19 @@
...
 
-COPY . /src
 WORKDIR /src
 
 ENV CGO_ENABLED=0
 
-RUN apk add git && ./build 
+RUN apk add git && \
+    git clone https://github.com/yggdrasil-network/yggdrasil-go.git . && \
+    ./build
```
jcgruenhage: *Why?*
It makes sure that you can only ever build the latest master commit and nothing else. This is horrible.

contrib/docker/Dockerfile
```diff
@@ -1,18 +1,19 @@
...
 
-FROM docker.io/alpine
+FROM alpine
```
jcgruenhage: See other comment above, same ambiguity problem.

contrib/docker/Dockerfile
```diff
@@ -1,18 +1,19 @@
...
 LABEL maintainer="Christer Waren/CWINFO <christer.waren@cwinfo.org>"
 
 COPY --from=builder /src/yggdrasil /usr/bin/yggdrasil
 COPY --from=builder /src/yggdrasilctl /usr/bin/yggdrasilctl
-COPY contrib/docker/entrypoint.sh /usr/bin/entrypoint.sh
+COPY --from=builder /src/contrib/docker/entrypoint.sh /usr/bin/entrypoint.sh
```
jcgruenhage: This would not be necessary if you hadn't done the clone thing.

PR #335, opened 2019-02-18
Version 0.3.3

neilalexander:
This is version 0.3.3. In terms of lines of code changed and the span of changes, this is probably one of our biggest pull requests yet - despite that it is mostly a maintenance/refactoring release and has not many user-facing changes.

Notable changes include:

- Node configuration is now stored centrally, rather than having fragments/copies distributed at startup time, which lets us implement...
- Dynamic reconfiguration, which allows reloading the configuration file to make changes during runtime by sending a `SIGHUP` signal (note: this only works with `-useconffile` and not `-useconf` and currently reconfiguring TUN/TAP is not supported)
- Support for building Yggdrasil as an iOS or Android framework if the appropriate tools (e.g. gomobile/gobind + SDKs) are available
- Flexible logging support, which allows for logging at different levels of verbosity
- Significant refactoring in various areas, including for link types (TCP, AWDL etc), generic streams and adapters
- Connection contexts used for TCP connections which allow more exotic socket options to be set, e.g.
  - Reusing the multicast socket across multiple running Yggdrasil instances
  - Allowing supported Macs to peer with other nearby Macs that aren't even on the same Wi-Fi network using AWDL
- Switch changes to improve parent selection
-  `ReadTimeout` option is now deprecated
- Simplified `systemd` service now in `contrib`
- macOS builds through CircleCI are now 64-bit only


Mikaela:
> * Dynamic reconfiguration, which allows reloading the configuration file to make changes during runtime by sending a `SIGHUP` signal (note: this only works with `-useconffile` and not `-useconf` and currently reconfiguring TUN/TAP is not supported)

Should `ExecReload=/bin/kill -HUP $MAINPID` added to `yggdrasil.service` in this case? If I am looking at the diff correctly, only mentioning to `yggdrasil-resume.service` was removed.

neilalexander:
Updated the `yggdrasil.service` file.

PR #348, opened 2019-03-01
contrib/busybox-init: add init.d script for busybox init

willeponken:
Hello there, I made this for my simple buildroot system, I though I'd add it here too :)

PR #349, opened 2019-03-01
contrib/busybox-init: add init.d script for busybox init

willeponken:
Heh, see: #348


neilalexander:
Thanks for contributing :-)

PR #350, opened 2019-03-01
Wake AWDL for multicast peering on macOS

neilalexander:
This wakes up AWDL for multicast peer discovery when the `awdl0` multicast interface is enabled. This allows Macs to directly peer with one another, even if they are on different Wi-Fi networks or not even connected to Wi-Fi at all.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@
 # Check https://circleci.com/docs/2.0/language-go/ for more details
 version: 2
 jobs:
-  build:
+  build-linux:
     docker:
       - image: circleci/golang:1.11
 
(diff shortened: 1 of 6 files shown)


neilalexander:
Marking as unfinished purely because I still need to figure out how to get CircleCI to handle these builds properly in a workflow.

@Arceliar - if you wouldn't mind reviewing the rest of the PR excluding the CircleCI stuff, that'd be great.

neilalexander:
OK - this now has a working parallel build pipeline, which uses both Linux and macOS CI workers.

This is available on the `yggdrasil-network` org within CircleCI as I requested the Seed plan there, but I don't know what this will do to forks. We may need to start using branches on the main repo instead if it causes us problems.

neilalexander:
That’s because the branch is looking for a status which is no longer being run in the new CI script. We need to update the protected branch rules to look for the new four entries and remove the original one.

PR #360, opened 2019-03-04
Support for multiple listeners

neilalexander:
This adds support for multiple listeners, e.g. more than one TCP listen address, or in theory mixed types of listeners (e.g. regular TCP sockets, UNIX domain sockets, etc):

1. The `Listen` configuration item is now an array. If it is currently a string then we attempt to convert it to an array either at startup or when running `-normaliseconf`
1. `tcpInterface` is now just `tcp`, in keeping with `awdl`/`awdlInterface`, and it now belongs to `link` rather than `Core`
1. `link` now has a `reconfigure` channel, and it cascades down to `tcp`, `awdl` etc.
1. `tcp` now supports holding and opening multiple listeners
1. The `Listen` configuration item can now be completely empty (e.g. not listening on any globally routable addresses) without affecting link-local multicast (see below)
1. `multicast.go` has been adapted so that it creates new TCP listeners for each interface (based on found link-local addresses) as this allows multicasting to work properly even when `Listen` addresses are not wildcard
1. Sets a static multicast interval of 15 seconds and reevaluates the available multicast interfaces upon each interval
1. Cleans up multicast listeners for interfaces that are no longer up or eligible (e.g. no multicast flag)

### DIFF
--- a/src/yggdrasil/multicast.go
+++ b/src/yggdrasil/multicast.go
@@ -149,11 +149,11 @@ func (m *multicast) announce() {
 				if l, ok := m.listeners[iface.Name]; !ok || l.listener == nil {
 					// No listener was found - let's create one
 					listenaddr := fmt.Sprintf("[%s%%%s]:0", addrIP, iface.Name)
-					if l, err := m.core.link.tcp.listen(listenaddr); err == nil {
+					if li, err := m.core.link.tcp.listen(listenaddr); err == nil {
 						m.core.log.Debugln("Started multicasting on", iface.Name)
 						// Store the listener so that we can stop it later if needed
-						m.listeners[iface.Name] = l
-						listener = l
+						m.listeners[iface.Name] = li
+						listener = li
 					}
 				} else {
 					// An existing listener was found
--- a/src/yggdrasil/multicast.go
+++ b/src/yggdrasil/multicast.go
@@ -27,7 +27,7 @@ func (m *multicast) init(core *Core) {
 		for {
 			e := <-m.reconfigure
 			m.myAddrMutex.Lock()
-			m.myAddr = m.core.tcp.getAddr()
+			m.myAddr = m.core.link.tcp.getAddr()
 			m.myAddrMutex.Unlock()
 			e <- nil
 		}
@@ -109,7 +109,7 @@ func (m *multicast) interfaces() []net.Interface {
 func (m *multicast) announce() {
 	var anAddr net.TCPAddr
 	m.myAddrMutex.Lock()
-	m.myAddr = m.core.tcp.getAddr()
+	m.myAddr = m.core.link.tcp.getAddr()
 	m.myAddrMutex.Unlock()
 	groupAddr, err := net.ResolveUDPAddr("udp6", m.groupAddr)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


src/yggdrasil/multicast.go
```diff
@@ -16,19 +15,16 @@ type multicast struct {
...
 	groupAddr   string
-	myAddr      *net.TCPAddr
-	myAddrMutex sync.RWMutex
+	listeners   map[string]*tcpListener
 }
 
 func (m *multicast) init(core *Core) {
 	m.core = core
 	m.reconfigure = make(chan chan error, 1)
+	m.listeners = make(map[string]*tcpListener)
```
Arceliar: It's kind of ugly that we have listeners here and also in the link's tcp struct. I'm not sure there's a better alternative, but I don't really like it...

src/yggdrasil/tcp.go
```diff
@@ -62,104 +56,145 @@ func (iface *tcpInterface) setExtraOptions(c net.Conn) {
...
 		select {
-		case <-iface.stop:
-			iface.core.log.Errorln("Stopping listener")
-			return
-		default:
+		case <-accepted:
 			if err != nil {
-				panic(err)
+				t.link.core.log.Errorln("Failed to accept connection:", err)
+				return
```
Arceliar: We should do the same cleanup (from the `l.stop` case) if we return for any reason. I guess maybe move that to a `defer` right below this loop starts, and then both `l.stop` and this `err` case could be a line or so of logging + a `return`, unless you have a better idea.

src/yggdrasil/multicast.go
neilalexander: The main reason for this is that the map in `tcp.go` is a `map[listenaddr]listener` whereas the map in `multicast.go` is a `map[interfacename]listener`. The duplication isn't ideal but since `tcp.go` and `multicast.go` are responsible for populating their own maps, I don't *believe* this should be a problem.

src/yggdrasil/tcp.go
```diff
@@ -267,25 +303,25 @@ func (iface *tcpInterface) call(saddr string, socksaddr *string, sintf string) {
...
-	iface.setExtraOptions(sock)
+	t.setExtraOptions(sock)
 	stream := stream{}
 	stream.init(sock)
 	local, _, _ := net.SplitHostPort(sock.LocalAddr().String())
 	remote, _, _ := net.SplitHostPort(sock.RemoteAddr().String())
 	remotelinklocal := net.ParseIP(remote).IsLinkLocalUnicast()
 	name := "tcp://" + sock.RemoteAddr().String()
-	link, err := iface.core.link.create(&stream, name, "tcp", local, remote, incoming, remotelinklocal)
+	link, err := t.link.core.link.create(&stream, name, "tcp", local, remote, incoming, remotelinklocal)
```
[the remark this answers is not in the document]
neilalexander: I've done this so that `proto` set to to `"socks"` (as that is used in output elsewhere), but we instead now resolve the true address of the peer and the proxy and put a full `socks://` connection string into the `name` field. The admin socket now returns the `linkInfo` `name` field instead of `remote` so that the full connect string is visible.

PR #373, opened 2019-03-09
Add AppArmor profile

l-n-s:
This PR adds a profile which can be used to harden yggdrasil setup with a Linux kernel security module [AppArmor](https://en.wikipedia.org/wiki/AppArmor).


neilalexander:
This looks fine to my untrained eyes. 

@Arceliar - happy with this contrib?

PR #382, opened 2019-03-13
Multicast fix

Arceliar:
Previously, if an interface remained up but an address was changed, then we would continue to keep the old listener open and advertise its old listen address. This fails if, for example, a node assigns new random link-local IPv6 addresses when switching wifi networks.

The change indexes listeners by the string representation of a `net.IPAddr`, which includes a `Zone` field with the interface name, instead of trying to index by interface name alone.

The catch is that, if you have more than one link-local address, then it may open and advertise more than one listener per interface, which could lead to duplicate connections. I'm not sure what the best way is to get around that... I figured I should commit what I have in case @neilalexander sees an obvious way to fix the multiple address issue (something like a `map[iface.Name]map[address]listener` maybe...).

### DIFF
--- a/src/yggdrasil/multicast.go
+++ b/src/yggdrasil/multicast.go
@@ -118,23 +118,13 @@ func (m *multicast) announce() {
 		panic(err)
 	}
 	for {
+		// Get the allowed interfaces and any matching link-local addresses
 		interfaces := m.interfaces()
-		// There might be interfaces that we configured listeners for but are no
-		// longer up - if that's the case then we should stop the listeners
-		for name, listener := range m.listeners {
-			if _, ok := interfaces[name]; !ok {
-				listener.stop <- true
-				delete(m.listeners, name)
-				m.core.log.Debugln("No longer multicasting on", name)
-			}
-		}
-		// Now that we have a list of valid interfaces from the operating system,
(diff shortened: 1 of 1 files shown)


src/yggdrasil/multicast.go
```diff
@@ -146,38 +136,55 @@ func (m *multicast) announce() {
...
-					continue
-				}
-				// Get the listener details and construct the multicast beacon
-				lladdr := listener.listener.Addr().String()
-				if a, err := net.ResolveTCPAddr("tcp6", lladdr); err == nil {
-					a.Zone = ""
-					destAddr.Zone = iface.Name
-					msg := []byte(a.String())
-					m.sock.WriteTo(msg, nil, destAddr)
+				addresses[ipAddr.String()] = ipAddr
```
Arceliar: I'm pretty sure that adding a `break` after this line would do the right thing in the case where there are multiple link-local addresses per link, *if* `iface.Addrs()` returns a slice that's sorted from oldest to newest. If not, then we could sort it ourselves, but weird things might happen if we add a new earlier address while keeping the old later one (listener closes, but connections don't, and then we could get a duplicate connection with the new listener).

neilalexander:
Superseded by #383.

PR #386, opened 2019-03-14
More docker support

yangm97:


neilalexander:
Please create PRs against the `develop` branch instead of `master`.

yangm97:
Done.

Arceliar:
Reopening so we can think about what to do here.

Compared to the file under `contrib/docker` (now under `contrib/docker/alpine`, this uses `-autoconf`, which means you get a different IP every time you start it, no clearnet peers, etc... I'm not sure how useful that is as the default docker file. But as I've said, I have almost 0 experience with docker, so I'm not the target audience for this.

yangm97:
@Arceliar the CMD is overridable. When you `docker run` without arguments, it will take the default CMD, but if the user mounts a config file and then adds -useconf... it will run from that instead.

yangm97:
I could add some concrete examples somewhere, maybe?

PR #395, opened 2019-03-24
Handle admin socket error cases better

### LINKED ISSUE #385: Invalid commands in yggdrasilctl return a generic error
```
$ sudo yggdrasilctl fsadkfjsk
Admin socket returned an error but didn't specify any error text
```

neilalexander:
This contains a couple of minor fixes to the admin socket:

- Return a more useful error when the client requests a command that doesn't exist (fixes #385)
- Reduce the logging level of admin socket recovers to `debug` instead of `error`

### DIFF
--- a/src/admin/admin.go
+++ b/src/admin/admin.go
@@ -381,11 +381,11 @@ func (a *AdminSocket) handleRequest(conn net.Conn) {
 		if r != nil {
 			send = Info{
 				"status": "error",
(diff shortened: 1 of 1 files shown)


neilalexander:
Deliberately not merging this yet as I've sent this PR and the webhook IDs to CircleCI, since this is yet another PR that didn't seem to trigger a CI build.

PR #399, opened 2019-03-27
Peer lookup by SRV or TXT records

neilalexander:
This adds support for two types of URLs:

- `srv://domain.com` (where it looks up SRV record `_yggdrasil._tcp.domain.com`)
- `txt://domain.com` (where it looks up TXT record `_yggdrasil.domain.com` where the TXT record contains a normal peer URI, e.g. `tcp://a.b.c.d:e`)

I'm not entirely sure I'm happy with this PR yet. It doesn't really respect priority fields or anything like that – it adds only the first record found in a SRV case to the peer list, although this *should* be the highest priority record. It doesn't attempt to fail over to lesser priority nodes, as we don't really have the machinery in place to allow this.

It does add a new dimension of configuration into `core.go` so that `srv://` and `txt://` peers are resolved only either at startup or during reconfiguration – I'm not sure if this is messy or can be done better. The problem it solves is that otherwise we'll hammer DNS with repeated requests as Go does not cache them.

That said, we do already probably hammer DNS a bit as it is if `tcp://` DNS peers are configured, particularly where the peers are not connected. We may want to look at that as well.

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -108,10 +108,13 @@ func readConfig(useconf *bool, useconffile *string, normaliseconf *bool) *nodeCo
 	// Check to see if the peers are in a parsable format, if not then default
 	// them to the TCP scheme
 	if peers, ok := dat["Peers"].([]interface{}); ok {
+	peerloop:
 		for index, peer := range peers {
 			uri := peer.(string)
-			if strings.HasPrefix(uri, "tcp://") || strings.HasPrefix(uri, "socks://") {
-				continue
+			for _, prefix := range []string{"tcp://", "socks://", "srv://", "txt://"} {
+				if strings.HasPrefix(uri, prefix) {
+					continue peerloop
+				}
 			}
 			if strings.HasPrefix(uri, "tcp:") {
 				uri = uri[4:]
@@ -121,11 +124,14 @@ func readConfig(useconf *bool, useconffile *string, normaliseconf *bool) *nodeCo
 	}
 	// Now do the same with the interface peers
 	if interfacepeers, ok := dat["InterfacePeers"].(map[string]interface{}); ok {
+	interfacepeerloop:
 		for intf, peers := range interfacepeers {
 			for index, peer := range peers.([]interface{}) {
 				uri := peer.(string)
-				if strings.HasPrefix(uri, "tcp://") || strings.HasPrefix(uri, "socks://") {
-					continue
+				for _, prefix := range []string{"tcp://", "socks://", "srv://", "txt://"} {
+					if strings.HasPrefix(uri, prefix) {
+						continue interfacepeerloop
+					}
 				}
 				if strings.HasPrefix(uri, "tcp:") {
 					uri = uri[4:]
--- a/src/yggdrasil/admin.go
+++ b/src/yggdrasil/admin.go
@@ -591,7 +591,11 @@ func (a *admin) printInfos(infos []admin_nodeInfo) string {
 
 // addPeer triggers a connection attempt to a node.
 func (a *admin) addPeer(addr string, sintf string) error {
-	err := a.core.link.call(addr, sintf)
+	entry, err := a.core.resolvePeerEntry(addr, sintf)
+	if err != nil {
+		return err
+	}
+	err = a.core.link.call(entry.uri, entry.sintf)
 	if err != nil {
 		return err
 	}
--- a/src/yggdrasil/core.go
+++ b/src/yggdrasil/core.go
@@ -404,12 +404,14 @@ func (c *Core) resolvePeerEntry(uri, sintf string) (peerEntry, error) {
 		if records, err := net.LookupTXT(recordname); err == nil {
 			for _, record := range records {
 				c.log.Traceln("Found TXT record:", record)
-				if !strings.HasPrefix(record, "txt://") {
-					return peerEntry{
-						uri:   fmt.Sprintf("%s", record),
-						sintf: sintf,
-					}, nil
+				if strings.HasPrefix(record, "txt://") {
+					// Prevents an infinite loop
+					continue
 				}
+				return peerEntry{
+					uri:   fmt.Sprintf("%s", record),
+					sintf: sintf,
+				}, nil
 			}
 		} else {
 			c.log.Debugln("TXT lookup failed:", err)
--- a/src/yggdrasil/core.go
+++ b/src/yggdrasil/core.go
@@ -2,8 +2,11 @@ package yggdrasil
 
 import (
 	"encoding/hex"
+	"fmt"
 	"io/ioutil"
 	"net"
+	"net/url"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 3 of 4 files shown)


Vort:
At least Windows and Ubuntu have built-in caches, so it is not a problem there.
But anyway, if caching is a problem, why not to make a cache?

Mikaela:
I think caching as long as the TTL says is better than relying on OS caching.

I don't know about Ubuntu nowadays, but when they moved to dnsmasq-base they were explicitly disabling caching, systemd-resolved may have it by default, but I have heard that Ubuntu has again moved to something new for network configuration and possibly DNS.

Then there are many users who have disabled systemd-resolved or prefer to manage their resolv.conf by themselves which disabled caching (and other features systemd-resolved may have enabled such as DNSSEC and opportunistic DNS over TLS).

Mikaela:
> srv://domain.com (where it looks up SRV record _yggdrasil._tcp.domain.com)

Does this have the same syntax as TXT?

> txt://domain.com (where it looks up TXT record _yggdrasil.domain.com where the TXT record contains a normal peer URI, e.g. tcp://a.b.c.d:e)

As the record types are the same, it's probably not OK to use CNAME or similar so people opting to use this will need to decide which one to use?

neilalexander:
> I think caching as long as the TTL says is better than relying on OS caching.

I am not convinced that Go respects OS caching settings anyway.

> Does this have the same syntax as TXT?

No, because SRV records do not allow free-flow input. They have set fields with set formats.

> As the record types are the same, it's probably not OK to use CNAME or similar so people opting to use this will need to decide which one to use?

Honestly I am not a fan of the current TXT approach – I added it just to prove that it would work.

I think I am more likely to change it so a SRV+TXT combination is used whereby the SRV record points to the endpoint and the TXT record *optionally* exists to contain additional data like a pinned public key. That way it would comply with the DNS-SD standard and be more easily reusable for multicast DNS.

cwchristerw:
> I think I am more likely to change it so a SRV+TXT combination is used whereby the SRV record points to the endpoint and the TXT record _optionally_ exists to contain additional data like a pinned public key. That way it would comply with the DNS-SD standard and be more easily reusable for multicast DNS.
+1 TO THIS SOLUTION

Also this needs to support multiple SRV records because standard does support setting multiple servers because there is priority & weight tags.

SRV DNS RECORD EXAMPLE
```
_service._proto.name. TTL class SRV priority weight port target.
_sip._tcp.example.com. 86400 IN SRV 0 5 5060 sipserver.example.com.
```

Do I add to peer list in future
`tcp://y.cwinfo.net`

DNS
```
_yggdrasil._tcp.y.cwinfo.net 3600 IN SRV 0 0 64982 cloudberry.fr1.servers.devices.cwinfo.net
_yggdrasil._tcp.y.cwinfo.net 3600 IN SRV 0 0 62506 fr2.servers.devices.cwinfo.net
_yggdrasil._tcp.y.cwinfo.net 3600 IN SRV 0 0 37145 de1.servers.devices.cwinfo.net
_yggdrasil._tcp.y.cwinfo.net 3600 IN SRV 0 0 37145 pl1.servers.devices.cwinfo.net

```

MORE COMING COMPUTER IS FREEZING

neilalexander:
> Also this needs to support multiple SRV records because standard does support setting multiple servers because there is priority & weight tags.

The problem here is that Yggdrasil doesn't have a concept of priority *or* weight for peers, so our options right now are:

- Add the first record (what the PR currently does) which happens to be the highest priority record
- Add a random record and hope that it is available
- Add every single record returned to us

cwchristerw:
> > Also this needs to support multiple SRV records because standard does support setting multiple servers because there is priority & weight tags.

> The problem here is that Yggdrasil doesn't have a concept of priority _or_ weight for peers, so our options right now are:
I know I have suggestion for country code based subdomains but when public peer provider doesn't provide public peer other countries it's possible that just some domain is enough. And Yggdrasil should have concept of priority or weight in dns srv because protocol it self chooses the best default route.

>     * Add the first record (what the PR currently does) which happens to be the highest priority record
In this case one of all in same priority should be selected.

>     * Add a random record and hope that it is available (not great)
DNS gives always random records if it has multiple in same srv.

>     * Add every single record returned to us (which will have absolutely no location awareness and won't be optimal in your example of geographically distinct peers)
This could be solved with simple subdomain or weight to country code translation but in current times  node it self doesn't know country that it's in.

Could be solved with this kind of solution or similar records...

## SUBDOMAIN EXAMPLES
(Or similar that includes country code in domain, but format be decided by each public node provider and there should be way to provide one address to connect to all public nodes in that domain for other public peers example, public peers should be interconnected.)

### EUROPE
`tcp://eu.y.cwinfo.net`
_yggdrasil._tcp.eu.y.cwinfo.net 3600 IN SRV 0 0 64982 cloudberry.fr1.servers.devices.cwinfo.net
_yggdrasil._tcp.eu.y.cwinfo.net 3600 IN SRV 0 0 62506 fr2.servers.devices.cwinfo.net
_yggdrasil._tcp.eu.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.eu.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.eu.y.cwinfo.net 3600 IN SRV 0 0 39593 de1.servers.devices.cwinfo.net
_yggdrasil._tcp.eu.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.eu.y.cwinfo.net 3600 IN SRV 0 0 37145 pl1.servers.devices.cwinfo.net
_yggdrasil._tcp.eu.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}

### FRANCE
CONFIG
`tcp://fr.y.cwinfo.net`

DNS
_yggdrasil._tcp.fr.y.cwinfo.net 3600 IN SRV 0 0 64982 cloudberry.fr1.servers.devices.cwinfo.net
_yggdrasil._tcp.fr.y.cwinfo.net 3600 IN SRV 0 0 62506 fr2.servers.devices.cwinfo.net
_yggdrasil._tcp.fr.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.fr.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}

### GERMANY
CONFIG
`tcp://de.y.cwinfo.net`

DNS
_yggdrasil._tcp.de.y.cwinfo.net 3600 IN SRV 0 0 39593 de1.servers.devices.cwinfo.net
_yggdrasil._tcp.de.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}

### POLAND
CONFIG
`tcp://pl.y.cwinfo.net`

DNS
_yggdrasil._tcp.pl.y.cwinfo.net 3600 IN SRV 0 0 37145 pl1.servers.devices.cwinfo.net
_yggdrasil._tcp.pl.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}


# FUTURE
### EARTH
`tcp://earth.y.cwinfo.net`
_yggdrasil._tcp.earth.y.cwinfo.net 3600 IN SRV 0 0 64982 cloudberry.fr1.servers.devices.cwinfo.net
_yggdrasil._tcp.earth.y.cwinfo.net 3600 IN SRV 0 0 62506 fr2.servers.devices.cwinfo.net
_yggdrasil._tcp.earth.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.earth.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.earth.y.cwinfo.net 3600 IN SRV 0 0 39593 de1.servers.devices.cwinfo.net
_yggdrasil._tcp.earth.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.earth.y.cwinfo.net 3600 IN SRV 0 0 37145 pl1.servers.devices.cwinfo.net
_yggdrasil._tcp.earth.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}

### MARS
`tcp://mars.y.cwinfo.net`
_yggdrasil._tcp.mars.y.cwinfo.net 3600 IN SRV 0 0 64982 cloudberry.fr1.servers.devices.cwinfo.net
_yggdrasil._tcp.mars.y.cwinfo.net 3600 IN SRV 0 0 62506 fr2.servers.devices.cwinfo.net
_yggdrasil._tcp.mars.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.mars.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.mars.y.cwinfo.net 3600 IN SRV 0 0 39593 de1.servers.devices.cwinfo.net
_yggdrasil._tcp.mars.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}
_yggdrasil._tcp.mars.y.cwinfo.net 3600 IN SRV 0 0 37145 pl1.servers.devices.cwinfo.net
_yggdrasil._tcp.mars.y.cwinfo.net 3600 IN TXT {PUBLIC KEY}

cwchristerw:
When connecting peer to itself does it brake anything if does this should be solved somehow.

cwchristerw:
If connection gets disconnected then should be checked again from dns to get new records if changed and test like described.

cwchristerw:
Proper way to use SRV records is to set tcp:// without port so it should resolve it from dns also it should ignore all A and AAAA in that subdomain and connect according SRV. In this way it's possible to serve website in same domain :smile: sorry for explaining avious things :smile:

cwchristerw:
> Honestly I am not a fan of the current TXT approach – I added it just to prove that it would work.
If need support solution without SRV records (all DNS services dont support). It's possible to make if no SRV then using different TXT format like, i think then it's also possible to support multiple peers in single domain.
mars.y.cwinfo.net 3600 IN TXT "tcp://a.b.c.d:e"
mars.y.cwinfo.net 3600 IN TXT "tcp://a.b.c.d.e.f.g.h:i"

cwchristerw:
Is there anything new about this feature, it would be useful when configuring desktops and laptops to use.

neilalexander:
This is stale and had a number of problems.

PR #401, opened 2019-03-28
Refactoring for Yggdrasil library

neilalexander:
This PR is a refactoring exercise.

So far it moves TUN/TAP and ICMPv6 into a separate package, and updates all the multivarious references. It also adds a new `NodeState` type for tracking current/previous node config and protecting them with a mutex.

More to come...


neilalexander:
OK so this PR grew arms and legs a bit. Rough list of changes:

- `tuntap`, `mobile`, `multicast` and `dummy` are now separate packages
- Various functions have been exported from various places in order to stitch this together and some exported functions have been renamed to make them more Go-library-like
- Node config is now returned from `Core.Start` in a `NodeState` which contains a mutex – it is expected to give this to other modules like `tuntap` or `dummy` so that we don't need to export from `Core` unnecessarily
- Generic dummy interface type was added to replace the old mobile send/recv functions – this is needed for iOS though since `gomobile` has no concept of channels
- Some admin functions are commented out because I haven't got hooks for those yet
- Yggdrasil can now start without an adapter (although obviously this isn't very useful) and needs to be supplied with an adapter interface before startup
- Rejected packets (e.g. from PMTUD) are now sent back through a "Rejected" channel so that the adapter can do the right thing (like ICMPv6 PTB with TUN/TAP) or ignore as appropriate
- Various godoc updates
- Debug builds are really broken – sorry!

As best as I can tell, apart from debug builds, everything still works and I've been able to get iOS builds working again.

There are still plenty of things to be done but I think it's best that we freeze this PR here and review i, tweak it, merge it etc so that we aren't doing way too much in one go.

@Arceliar – feel free to push any changes as you see fit and/or merge when ready.

neilalexander:
@Arceliar Did you get the chance to go over this thoroughly? Reluctant to make any more changes until this PR is fine.

.circleci/config.yml
```diff
@@ -24,11 +24,11 @@ jobs:
...
 
-      - run:
-          name: Test debug builds
-          command: |
-              ./build -d
-              test -f yggdrasil && test -f yggdrasilctl
+    #  - run:
+    #      name: Test debug builds
+    #      command: |
+    #          ./build -d
```
Arceliar: reminder to re-enable after we fix `debug.go` in a later PR

cmd/yggdrasil/main.go
```diff
@@ -244,22 +248,29 @@ func main() {
 	// Setup the Yggdrasil node itself. The node{} type includes a Core, so we
 	// don't need to create this manually.
 	n := node{}
-	// Now that we have a working configuration, we can now actually start
-	// Yggdrasil. This will start the router, switch, DHT node, TCP and UDP
-	// sockets, TUN/TAP adapter and multicast discovery port.
-	if err := n.core.Start(cfg, logger); err != nil {
+	// Before we start the node, set the TUN/TAP to be our router adapter
+	n.core.SetRouterAdapter(&n.tuntap)
```
Arceliar: I still feel like this is backwards. The core/router should have in/out channels or functions (functions is probably better, since you can't accidentally overwrite them), and then have the tun/tap make use of those, instead of injecting a dependency after the fact. But that requires a little care to make sure that whatever the `Read()` equivalent is doesn't cause the core to either block or drop traffic unnecessarily if it's not being called often enough. So I'm not sure we should try to deal with it now. Just a reminder to take another look at this later, since it's ultimately how the user is going to interact and send/receive traffic.

src/config/config.go
```diff
@@ -2,11 +2,36 @@ package config
...
+
+// Get returns both the current and previous node configs
+func (s *NodeState) Get() (NodeConfig, NodeConfig) {
+	s.Mutex.RLock()
+	defer s.Mutex.RUnlock()
+	return s.Current, s.Previous
+}
+
+// Replace the node configuration with new configuration
+func (s *NodeState) Replace(n NodeConfig) NodeConfig {
```
Arceliar: If you call this and then call `Get()` to get the old config again, you potentially have a race. You could get around the race by returning the current and previous config when you call this (or just the previous, since the caller will already have the new config). This is (theoretically) an serious problem since `Replace` could be called again before all relevant goroutines have all called `Get` and made any necessary changes. In practice, this shouldn't normally be an issue, but it would be an absolute nightmare to debug if e.g. a node is under heavy load and gets another `SIGHUP` before it finishes updating everything.

src/mobile/awdl.go
```diff
@@ -1,5 +1,6 @@
-package yggdrasil
+package mobile
 
+/*
```
Arceliar: Reminder to uncomment and fix this, or else just give up on it entirely if apple isn't going to add support for what we need.

src/mobile/mobile_ios.go
```diff
@@ -29,6 +26,7 @@ func (nsl MobileLogger) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
+/*
```
Arceliar: Another thing to remember to uncomment/fix or else give up on, depending on if apple ever does anything to improve the situation.

src/tuntap/tun.go
```diff
@@ -0,0 +1,361 @@
+package tuntap
```
Arceliar: I still don't like that this file is "new" instead of recognized as a move, but I guess there's no point in holding things up longer just to try and fix that.

Nothing looks obviously wrong, but since it's not a diff, I can't guarantee that I noticed any nontrivial changes.

src/yggdrasil/adapter.go
```diff
@@ -1,18 +1,47 @@
...
+	"github.com/gologme/log"
+	"github.com/yggdrasil-network/yggdrasil-go/src/address"
+	"github.com/yggdrasil-network/yggdrasil-go/src/config"
+)
+
+// Adapter defines the minimum required struct members for an adapter type. This
+// is now the base type for adapters like tun.go. When implementing a new
+// adapter type, you should extend the adapter struct with this one and should
+// call the Adapter.Init() function when initialising.
 type Adapter struct {
```
Arceliar: General reminder that I'm not really happy with this interface and want to do something else eventually.

For example, we might want to have `src/tuntap` (or some utility library that it uses) figure out the NodeID to use based on the packet's , then do something like a `Dial()` but with Yggdrasil's NodeIDs, and get back something resembling a `net.Conn`, or maybe skip the `Dial()` step and just expose a `net.PacketConn`. The latter may be easier, but keeps a lot of session management logic in `src/yggdrasil` that may make more sense elsewhere. The point is, ideally we should try to conform to `net`'s interfaces, so we can be a drop-in replacement for other things.

src/yggdrasil/ckr.go
```diff
@@ -55,25 +55,24 @@ func (c *cryptokey) init(core *Core) {
 // Configure the CKR routes - this must only ever be called from the router
```
Arceliar: In principle, this should also get refactored away (maybe as part of `src/tuntap`, or maybe a separate library that `tuntap` uses). Not something to do now, just a reminder to look at it in the next phase of refactoring.

src/yggdrasil/core.go
```diff
@@ -121,15 +111,13 @@ func (c *Core) addPeerLoop() {
...
+// config.NodeConfig and then signals the various module goroutines to
+// reconfigure themselves if needed.
 func (c *Core) UpdateConfig(config *config.NodeConfig) {
 	c.log.Infoln("Reloading configuration...")
 
-	c.configMutex.Lock()
-	c.configOld = c.config
-	c.config = *config
-	c.configMutex.Unlock()
+	c.config.Replace(*config)
```
Arceliar: Just a reminder that calling `Get` later is going to race, so it would be better to get the old config as a return arg, and then pass the old/new configs into the `reconfigure` channels, so other components know *exactly* what they need to compare.

src/config/config.go
neilalexander: I've updated `Replace` to return both the current and previous configs now. I wasn't really sure of a better way to handle this.

src/yggdrasil/adapter.go
neilalexander: Yep – I'd also like to get to the point where `Adapter` is not really necessary, but for now it solved a short-term problem just to keep Yggdrasil buildable and usable with this PR.

neilalexander:
The review has quite a lot of to-dos to take forward – shall we merge for now to get some more `develop` testers on-board and start to fix them in the next PR?

PR #410, opened 2019-04-07
admin-api: move from tcp/9001 to tcp/3959

wfleurant:
tcp/9001 conflicts with Tor ORPort.  
tcp/3959 has service named 'treehopper' in nmap-services. *\*shrug\**


neilalexander:
Which platform are you running that triggers the `other` defaults out of curiosity?

If it’s a platform that supports UNIX sockets then we should add a defaults file for it to default to those.

Mikaela:
> tcp/9001 conflicts with Tor ORPort.

I think they encourage changing the default anyway so it's not so easy to block. Personally I have put it to 443 so it's less likely to get blocked by some strict firewalls.

> Unless Tor is listening on UDP/9001, it’s also unnecessary.

Tor is TCP-only and I don't think it's listening on UDP. I think 9001/TCP/Orport is also disabled by default unless you become a relay.

wfleurant:
Caracho! I reverted my multicast mixup back to udp/9001.  
For now I've been protecting the admin-api with firewall rules.  
But since I kept running into tor & yggdrasil fighting for that port  
I wanted to change it upstream.  

If the the admin-api was available to specific cryptokeys  
I'd bind it to the ygg6 address at tcp/9001 and be done.

Another perhaps admin-api request is for multiple admin-api  
endpoints. I'd use by IP and locally with unix socket.  

No worries if this PR goes nowhere -- TBH I'd be happy if  
more people hit the same ports up because they'd be running  
both yggdrasil and tor-relays :)

PR #416, opened 2019-05-11
contribute decently fast yggdrasil address generator in C

fifteenthcommotion:
I think the yggdrasil-brute-simple address generator is now mature.
https://github.com/fifteenthcommotion/yggdrasil-brute-simple

### DIFF
--- a/contrib/yggdrasil-brute-simple/.gitignore
+++ b/contrib/yggdrasil-brute-simple/.gitignore
@@ -1,0 +1,7 @@
+*
+!.gitignore
+!Makefile
+!README.md
+!LICENSE
+!*.c
+!*.h
--- a/contrib/yggdrasil-brute-simple/.gitignore
+++ b/contrib/yggdrasil-brute-simple/.gitignore
@@ -1,7 +1,0 @@
-*
-!.gitignore
-!Makefile
-!README.md
-!LICENSE
-!*.c
-!*.h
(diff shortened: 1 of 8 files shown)


contrib/yggdrasil-brute-simple/.gitignore
```diff
@@ -0,0 +1,7 @@
+*
```
Mikaela: I wonder if you intend this directory to be a submodule instead? https://git-scm.com/book/en/v2/Git-Tools-Submodules

contrib/yggdrasil-brute-simple/LICENSE
```diff
@@ -0,0 +1 @@
+As public domain as possible. Don't blame me if your computer explodes.
```
Mikaela: I think CC0 might be more legal way to say this. https://creativecommons.org/choose/zero/

fifteenthcommotion: Apparently CC0 has some problems with explicitly not waiving patent rights.
https://opensource.stackexchange.com/questions/133/how-could-using-code-released-under-cc0-infringe-on-the-authors-patents

I'll probably use the Unlicense as it seems to be fairly short boilerplate for releasing code into the public domain and disclaiming warranty.

I expect to rarely (hopefully never) have to modify this code and I would rather not complicate the process of `git clone`-ing.

fifteenthcommotion:
Licensing is kind of a mess so I decided to add the CC0 dedication as well. There are now half as many licensing lines as there are lines of code.

neilalexander:
Thanks very much for contributing this!

PR #419, opened 2019-05-17
Modular Yggdrasil

neilalexander:
Although there are very few user-facing changes for regular Yggdrasil users, this is a massive PR which encompasses the following:

- The effective creation of a lightweight `libyggdrasil`, where Yggdrasil can be used as an application transport using Go `net.Conn`/`net.Listener`-style interfaces, similar to regular sockets, without unnecessary code
- A substantial amount of modularisation of the Yggdrasil codebase, including breaking out the admin socket, TUN/TAP, multicast, mobile etc into separate modules and rearranging the code path (e.g. moving all IP-specific routing logic and CKR into TUN/TAP)
- A re-fashioning of the `yggdrasil` binary to make use of the new interfaces

There are some performance reductions in this PR, although at this stage I am not sure that they are severe enough to be noticed by the average user. I just haven't really optimised the new code paths heavily yet - I am optimistic we can resolve that easily enough.

The new connection API looks something like this:

- `yggdrasil.Conn`, has `Read`, `Write`, `Close`, `LocalAddr`, `RemoteAddr`, `SetDeadline` etc functions
- `yggdrasil.Listener`, has `Accept`, `Addr` and `Close` functions
- `yggdrasil.Dialer`, has `Dial `and `DialByNodeIDandMask` functions

The API is still subject to change.


neilalexander:
Since you have a much better understanding of searches than I do, I’m wondering if you wouldn’t mind taking a look at that particular bit? 

I think the deadline stuff is much easier, we just need to work out some values that ultimately make sense.

PR #432, opened 2019-06-11
Implement session gatekeeper functions

neilalexander:
This adds a new API function `SetSessionGatekeeper` which takes a function of type `func(pubkey *crypto.BoxPubKey, initiator bool) bool`. 

The gatekeeper function is called each time a new session is initiated, either incoming or outgoing. The `pubkey` parameter contains the permanent public key of the remote host and the `initiator` function returns `true` if we were the ones to initiate the connection (false otherwise). The function should return `true` if it wants to allow the session or `false` if it wants to reject it.

In addition to this, the session firewall is now implemented in `cmd/yggdrasil` using this new API. This means that `session.go` is a bit leaner now and more complex behaviour can be implemented if required.

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/hex"
 	"encoding/json"
 	"flag"
 	"fmt"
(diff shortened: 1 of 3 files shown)


neilalexander:
I think we do already? One case is for incoming, one case is for outgoing (`true` vs `false`).
```
src/yggdrasil/session.go:       if !ss.isSessionAllowed(theirPermKey, true) {
src/yggdrasil/session.go:       if !isIn && !ss.isSessionAllowed(&ping.SendPermPub, false) {
```

PR #433, opened 2019-06-11
Try to build the new RPM using CircleCI

neilalexander:
This is an attempt to use the new yggdrasil-network/yggdrasil-package-rpm sources to build RPM files. 

Outstanding problems include:

- [ ] Not able to target other architectures other than the source RPM and `x86_64` because cross-compiling using `rpmbuild --target` tries to pepper things with 32-bit libraries that aren't installed
- [x] The `Conflicts` statement isn't updated at any point to reflect whether it conflicts with `yggdrasil`, `yggdrasil-develop` or both
- [x] The package version number for `develop` packages has a `.` instead of a `-` because, well, RPM


neilalexander:
I don't think we can do anything about crossbuilding the RPMs for other platforms but we have COPR builds to help us with that. In any case this is better than the horribly broken Alien RPMs.

PR #458, opened 2019-07-09
Correcting typo in headline

reuank:


neilalexander:
I've changed the target branch to `develop` as this is where all PRs are sent to until we publish a release.

Thanks for this though—rather embarrassing mistake!

PR #473, opened 2019-07-26
Possible fix for macos self-traffic

Arceliar:
Apparently macOS can't reach itself via its own yggdrasil address, presumably because it sends traffic to the tun instead of using a loopback.

My best guess as to where the traffic gets lost is in the search / DHT lookup step. This change causes the search to initialize with just our own node, instead of the contents of our own DHT. That should force an ordinary DHT lookup to happen, which should fix the search problem, if indeed that turns out to be where the macOS self-traffic gets lost.

Needs testing by someone with a mac.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,11 +32,11 @@ jobs:
               sudo apt-get install -y rpm file
               mkdir -p ~/rpmbuild/BUILD ~/rpmbuild/RPMS ~/rpmbuild/SOURCES ~/rpmbuild/SPECS ~/rpmbuild/SRPMS
 
-    #  - run:
-    #      name: Test debug builds
-    #      command: |
-    #          ./build -d
-    #          test -f yggdrasil && test -f yggdrasilctl
+      - run:
+          name: Test debug builds
+          command: |
+              ./build -d
+              test -f yggdrasil && test -f yggdrasilctl
 
       - run:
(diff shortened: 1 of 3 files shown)


neilalexander:
I've added a couple of flags to the interface address which I think are probably good judging from other interfaces on my system but sadly they do not fix the problem. Let's merge these anyway for good measure.

I don't really think this is an Yggdrasil-specific problem, rather a problem in how Darwin handles the loopback routes.

The routing table clearly shows an entry for the interface address via `lo0`:
```
200::/7                                 fe80::aabe:27ff:fec1:2520%utun0 Uc            utun0       
201:6243:84d5:813a:2a98:e6a:4279:318    link#19                         UHL             lo0
```
... and macOS thinks it should use it:
```
 ~  route get -inet6 201:6243:84d5:813a:2a98:e6a:4279:318
   route to: 201:6243:84d5:813a:2a98:e6a:4279:318
destination: 201:6243:84d5:813a:2a98:e6a:4279:318
  interface: lo0
      flags: <UP,HOST,DONE,LLINFO,LOCAL>
 recvpipe  sendpipe  ssthresh  rtt,msec    rttvar  hopcount      mtu     expire
       0         0         0         0         0         0     16384         0
```
... but then doesn't. I have no idea why, I assume this is a bug in macOS.

Arceliar:
Fixes at least some issues, so merging.

PR #497, opened 2019-08-10
Implement feature from https://github.com/yggdrasil-network/yggdrasil…

Slex:
I put version in a separate file, suggesting that in the future we may need migrations, for example, for configs.


neilalexander:
Thanks for this!

PR #512, opened 2019-08-18
Cryptokey routing changes

neilalexander:
This PR is designed to make CKR clearer, easier to use and more robust.

In summary:
- Configuration options are now renamed:
  - `IPv4Sources` is now `IPv4LocalSubnets`
  - `IPv4Destinations` is now `IPv4RemoteSubnets`
  - `IPv6Sources` is now `IPv6LocalSubnets`
  - `IPv6Destinations` is now `IPv6RemoteSubnets`
- Yggdrasil will automatically detect the old configuration item names when reading the config and `-normaliseconf` will rename them correctly to their new names
- Handling of CKR policies in the TUN/TAP adapter is now more robust
- A number of functions in `ckr.go` have been renamed
- Admin socket functions for CKR have been renamed to match the above

### DIFF
--- a/src/tuntap/iface.go
+++ b/src/tuntap/iface.go
@@ -21,6 +21,7 @@ func (tun *TunAdapter) writer() error {
 		if n == 0 {
 			continue
 		}
+		var srcAddr address.Address
 		var dstAddr address.Address
 		var addrlen int
 		// Check whether the packet is IPv4, IPv6 or neither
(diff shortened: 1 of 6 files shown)


src/tuntap/iface.go
```diff
@@ -21,25 +21,50 @@ func (tun *TunAdapter) writer() error {
...
+		} else {
+			// Neither IPv4 nor IPv6
+			return errors.New("Invalid address family")
+		}
+		// Check the crypto-key routing rules next
+		if tun.ckr.isEnabled() {
+			if !tun.ckr.isValidLocalAddress(dstAddr, addrlen) {
+				util.PutBytes(b)
+				continue
+			}
```
Arceliar: We should also check that the remote side's source address is one that we'll accept.

neilalexander: I've added in a check here that the source address matches *one of the ranges* we have configured as a remote, but without having any context for which key gave us this TUN/TAP traffic I can't improve the check to verify the specific key. Therefore I'm not sure how much value the check really gives us.

Arceliar:
Doesn't seem to want to run circleci build, but it seems to work fine locally, so I'm going to go ahead and merge.

Issue #487, opened 2019-08-06
Windows: Bad iperf3 results


Vort:
Driver: `9.9.2_3`
iPerf version: `3.1.3`
Yggdrasil version: `0.3.6-0026`
Command line: `iperf3 -6 -c 201:6779:c497:82f:68fd:188d:4338:1234`
OS: Windows 7 SP1 x64

Results are not just bad, they are totally wrong:
```
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-1.00   sec   256 KBytes  2.10 Mbits/sec
[  4]   1.00-2.00   sec  0.00 Bytes  0.00 bits/sec
[  4]   2.00-3.00   sec  0.00 Bytes  0.00 bits/sec
[  4]   3.00-4.00   sec  0.00 Bytes  0.00 bits/sec
[  4]   4.00-5.00   sec  0.00 Bytes  0.00 bits/sec
[  4]   5.00-6.00   sec  0.00 Bytes  0.00 bits/sec
[  4]   6.00-7.00   sec  0.00 Bytes  0.00 bits/sec
[  4]   7.00-8.00   sec  0.00 Bytes  0.00 bits/sec
[  4]   8.00-9.00   sec  0.00 Bytes  0.00 bits/sec
[  4]   9.00-10.00  sec  0.00 Bytes  0.00 bits/sec
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-10.00  sec   256 KBytes   210 Kbits/sec                  sender
[  4]   0.00-10.00  sec  64.2 KBytes  52.6 Kbits/sec                  receiver
```

My node is connected directly to `201:6779:c497:82f:68fd:188d:4338:1234` peer with ping of 45ms.
My Internet connection have 100 Mbit/s bandwidth.

I think this problem may be related to #472 (#456).

Vort:
One of the reasons for such bad result can be packet reordering.
It happens both during the sends and receives.
Most likely because of race conditions.

Test sequence: sending of 16 UDP packets with 1200 bytes size.
Tested version: `0.3.6-0043` (70a118ae98164e93ad122d187a18f1a0d0403b90)

![ygg_reorder](https://user-images.githubusercontent.com/1242858/62877444-bd62eb80-bd2f-11e9-8ac4-6bbf9fa308c8.png)

neilalexander:
[...]

PR #517, opened 2019-08-22
Minor changes to how tuntap dials a new conn

Arceliar:
make sure we buffer packets in order when waiting for a dial to finish, and make sure we can't leak the dial goroutine if the conn is closed before all buffered packets are sent

Possibly related to packet reordering in the first few packets of a stream, as shown in #487, but definitely shouldn't be related to any reordering later in the stream.

EDIT: Note that this is also solved in #519, and the two conflict, so if it's not urgent then we may want to just wait for that PR instead. Also, @Vort reported in https://github.com/yggdrasil-network/yggdrasil-go/issues/487#issuecomment-523783796 that the MTU needed to be reduced to 65521 on Windows to prevent packet loss, so we should perhaps update that too.

### DIFF
--- a/src/tuntap/iface.go
+++ b/src/tuntap/iface.go
@@ -203,16 +203,15 @@ func (tun *TunAdapter) readerPacketHandler(ch chan []byte) {
 				panic("Given empty dstNodeID and dstNodeIDMask - this shouldn't happen")
 			}
 			// Dial to the remote node
+			tun.mutex.Lock()
+			_, known := tun.dials[*dstNodeID]
+			tun.dials[*dstNodeID] = append(tun.dials[*dstNodeID], bs)
+			for len(tun.dials[*dstNodeID]) > 32 {
+				util.PutBytes(tun.dials[*dstNodeID][0])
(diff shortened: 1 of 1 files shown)


Vort:
Not sure if it is related, but let it be here:
```
2019/08/22 08:12:38 Stopping...
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x8 pc=0x5ef281]

goroutine 1 [running]:
github.com/yggdrasil-network/yggdrasil-go/src/tuntap.(*TunAdapter).Stop(...)
	E:/_Projects/_Test/yggdrasil-go/src/tuntap/tun.go:176
main.(*node).shutdown(0xc0000af500)
	E:/_Projects/_Test/yggdrasil-go/cmd/yggdrasil/main.go:297 +0xc1
main.main()
	E:/_Projects/_Test/yggdrasil-go/cmd/yggdrasil/main.go:291 +0xe1d
```

Arceliar:
Equivalent changes merged in #519, so closing this one.

PR #519, opened 2019-08-24
Actors

Arceliar:
This is an exercise to see if I can rewrite the main library's internals to stop using goroutines that communicate by channels, and start using an actor model approach instead. If it turns out to work, then we need to test and benchmark a lot to make sure it's not doing something stupid or leading to unexpected performance regressions.

Things I expect need a rewrite for this to be useful:

- [x] TunAdapter (a few parts, mainly ckr, still rely on mutexes for safety)
- [x] tunConn (replace workers, except for timeout check)
- [x] Conn (replace the *NoCopy interfaces while we're at it?)
- [x] sessionInfo (and replace workers)
- [x] router (replace mainLoop)
- [x] peer (replace linkLoop, what else?)
- [x] switchTable (replace worker)
- [x] link (needs a lot of testing, and no longer sends legacy keep-alive traffic every 4s, so not compatible with very old nodes -- we could re-add this, but now seems as good a time as any to drop it...)

The bulk of the rewrite is done at this point I believe. Now it's mostly testing/debugging, cleanup, and possibly rewriting some ugly parts (things were originally written with channels in mind, so some things probably make no sense in an actor model... e.g. things that use channel closes to signal).


Vort:
> EDIT: Note that this is also solved in #519

Did you made tests?
I see exactly the same behaviour as before.
To test it you can use iperf3 inside virtual machines with `-u` switch.
For me there are lots of these messages:
`iperf3: OUT OF ORDER - incoming packet = 12 and received packet = 30 AND SP = 5`

Vort:
Actually, single VM is enough for tests.
Here is command:
`iperf3 -c 201:6779:c497:82f:68fd:188d:4338:1234 -u`
And here is result:
ygg_iperf_ooo

src/yggdrasil/core.go
```diff
@@ -180,7 +186,9 @@ func (c *Core) Start(nc *config.NodeConfig, log *log.Logger) (*config.NodeState,
 
 	c.config.Mutex.RLock()
 	if c.config.Current.SwitchOptions.MaxTotalQueueSize >= SwitchQueueTotalMinSize {
-		c.switchTable.queueTotalMaxSize = c.config.Current.SwitchOptions.MaxTotalQueueSize
+		phony.Block(&c.switchTable, func() {
```
neilalexander: We might want to reverse this somewhat - that the switch takes the config mute and updates the `totalMaxSize` value itself when starting/reconfiguring, so that we don't need to do this here.

src/yggdrasil/debug.go
```diff
@@ -257,8 +259,8 @@ func DEBUG_wire_encode_coords(coords []byte) []byte {
 
 func (c *Core) DEBUG_getDHTSize() int {
```
neilalexander: If we are going to rewrite the sim, is this function needed? I suspect a lot of what's in `debug.go` isn't needed anymore.

src/yggdrasil/core.go
```diff
@@ -117,28 +112,39 @@ func (c *Core) addPeerLoop() {
...
+	for actor, functions := range components {
+		for _, function := range functions {
+			response := make(chan error)
+			phony.Block(actor, func() {
+				function(response)
+			})
+			for err := range response {
+				c.log.Errorln(err)
+				errors++
+			}
```
neilalexander: I've already popped a comment in here but I don't know that counting errors when reconfiguring the node actually really gives us anything. This is yet another use of `phony.Block` which we could probably turn into a `phony.Act` if we are happy to just let the actors get on with it and report/handle errors themselves.

src/yggdrasil/dht.go
```diff
@@ -65,33 +65,28 @@ type dhtReqKey struct {
...
-	reqs        map[dhtReqKey]time.Time          // Keeps track of recent outstanding requests
-	callbacks   map[dhtReqKey][]dht_callbackInfo // Search and admin lookup callbacks
+	router    *router
+	nodeID    crypto.NodeID
+	reqs      map[dhtReqKey]time.Time          // Keeps track of recent outstanding requests
+	callbacks map[dhtReqKey][]dht_callbackInfo // Search and admin lookup callbacks
 	// These next two could be replaced by a single linked list or similar...
 	table map[crypto.NodeID]*dhtInfo
 	imp   []*dhtInfo
 }
```
neilalexander: Is it better for `dht` to become an actor itself rather than depending on being in the router goroutine?

src/yggdrasil/router.go
```diff
@@ -62,120 +62,107 @@ func (r *router) init(core *Core) {
...
+	r.Act(from, func() {
+		r.sessions.reset()
+		r.dht.reset()
+	})
+}
+
+// TODO remove reconfigure so this is just a ticker loop
+// and then find something better than a ticker loop to schedule things...
+func (r *router) doMaintenance() {
+	phony.Block(r, func() {
```
neilalexander: Does this need to be blocking?

src/yggdrasil/core.go
```diff
@@ -117,28 +112,39 @@ func (c *Core) addPeerLoop() {
...
-	}
-
-	for _, component := range components {
-		response := make(chan error)
-		component <- response
-		if err := <-response; err != nil {
-			c.log.Errorln(err)
-			errors++
+	// Each reconfigure function should pass any errors to the channel, then close it
+	components := map[phony.Actor][]func(chan error){
```
neilalexander: This is about the least bad way I could think of to do this for now, I'm not sure if it's 100% correct but it seems to work okay in testing.

PR #542, opened 2019-09-10
Switch

Arceliar:
Something to try to stabilize switch behavior a bit.

When there are multiple idle links the same distance from the destination, send to the one which forwarded the most recent root update the fastest. That should tend towards selecting paths that give a low latency to the root, and also (hopefully) give relatively stable behavior, since root updates only happen so frequently.

If this does not work, we could also try only selecting the *best* link according to some metric, and buffer traffic if that link happens to be busy. That may be necessary if packets are coming in faster than we send them (particularly if TCP head-of-line blocking causes a bunch of packets to arrive all at once), since writes are probably inherently slower than reads.


Arceliar:
I think with that, with the added change to buffer traffic when the best peer is busy, this is about as good as things are going to get without significant changes to the switch. Which we should do, but I'd rather get the quick fixes into v0.3.9 and do that release first.

PR #548, opened 2019-09-18
Bugfixes

neilalexander:
This fixes a couple of minor bugs that I've found while testing today.

### DIFF
--- a/build
+++ b/build
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -ef
+
 PKGSRC=${PKGSRC:-github.com/yggdrasil-network/yggdrasil-go/src/version}
 PKGNAME=${PKGNAME:-$(sh contrib/semver/name.sh)}
 PKGVER=${PKGVER:-$(sh contrib/semver/version.sh --bare)}
--- a/build
+++ b/build
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-set -ef
-
 PKGSRC=${PKGSRC:-github.com/yggdrasil-network/yggdrasil-go/src/version}
 PKGNAME=${PKGNAME:-$(sh contrib/semver/name.sh)}
 PKGVER=${PKGVER:-$(sh contrib/semver/version.sh --bare)}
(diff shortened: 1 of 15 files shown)


build
```diff
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-set -ef
```
Arceliar: If the version/name scripts don't return non-zero exit codes anymore, then we can keep this I think.

cmd/yggdrasil/main.go
```diff
@@ -292,10 +292,10 @@ exit:
 }
 
 func (n *node) shutdown() {
-	n.core.Stop()
 	n.admin.Stop()
 	n.multicast.Stop()
 	n.tuntap.Stop()
+	n.core.Stop()
```
Arceliar: Order seems reasonable to me... as long as the core can still try to write to the tun/tap, and the tun/tap just fails when this happens.

cmd/yggdrasil/main.go
```diff
@@ -292,10 +292,10 @@ exit:
...
 	os.Exit(0)
```
Arceliar: We should get rid of all use of `os.Exit(0)`, just let things return (and run `defer`s) if there's no error.

src/yggdrasil/core.go
```diff
@@ -187,4 +191,12 @@ func (c *Core) Stop() {
 // This function is unsafe and should only be ran by the core actor.
 func (c *Core) _stop() {
 	c.log.Infoln("Stopping...")
+	if c.addPeerTimer != nil {
+		c.addPeerTimer.Stop()
+	}
+	c.link.stop()
+	for _, peer := range c.GetPeers() {
```
Arceliar: This races with the `c.link.stop()`, if there are any connections that have started (either still `Dial`ing, or successfully `Dial`ed or `Accept`ed but not done running the link handler) but which haven't created a peer yet by the time we get here.

Arceliar: Race should be fixed now, `c.link.stop()` will now signal links to exit before calling `tcp.stop()`, and `tcp.stop()` will block until all listeners have exited and all connections have closed (via a waitgroup). This could probably be done a little cleaner with a `util.Cancellation` but what I'd really like to do is get rid of the idle goroutines that sit around waiting for something to unblock, so we should come back to that later.

PR #551, opened 2019-09-20
Gradually increase multicast interval from startup

neilalexander:
Yggdrasil has always sent out multicast beacons every 15 seconds when enabled. This PR is a simple change to start that interval at 1s, and then add another 1s with each beacon until the interval reaches 15s.

This may help Yggdrasil to peer faster in some cases where Yggdrasil is brought up during system boot but some network interfaces may not be configured yet.

### DIFF
--- a/src/multicast/multicast.go
+++ b/src/multicast/multicast.go
@@ -26,12 +26,17 @@ type Multicast struct {
 	log             *log.Logger
 	sock            *ipv6.PacketConn
 	groupAddr       string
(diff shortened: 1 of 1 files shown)


Arceliar:
Slightly modified this. Now it has a separate interval per interface, and it checks all interfaces once per second to see if it needs to start/stop a listener or send a beacon. This should make things react more quickly when an interface is brought up (e.g. switching wifi on/off or connecting/disconnecting an ethernet cable).

PR #556, opened 2019-09-21
Switch hack

Arceliar:
So the trick here is, when we receive a packet and have an idle link, don't immediately send to that link. Instead, remove the link from the idle list (as normal), but buffer the packet and launch a new goroutine that will (at some point in the future, at the scheduler's whim) send a message to the switch notifying it of the idle peer. What I hope it does is give the best link a chance to forward that packet and notify the switch that it's idle before the 2nd best link's fake idle notification gets back to the switch. I'm not convinced that it actually does that, but whatever is really happening, it *seems* to help things somewhat in my tests.

Running some tests in network namespaces with artificial latency, this patch seems to help. If a node has 2 links with identical latency, it seems to result in no retransmissions (after the first second or so) in iperf3. If a node has two links with wildly different latency, then it flaps a lot, but it still seems to be more stable than without (e.g. it holds steady at the average bandwidth instead of oscillating wildly).

Also added some minor change to how TCP delays after redialing, since this was taking too long to make namespaces settle into a working state when I started them up at the same time and added delay to the veth devices.

### DIFF
--- a/src/yggdrasil/link.go
+++ b/src/yggdrasil/link.go
@@ -64,6 +64,8 @@ type linkInterface struct {
 	closeTimer     *time.Timer // Fires when the link has been idle so long we need to close it
 	inSwitch       bool        // True if the switch is tracking this link
 	stalled        bool        // True if we haven't been receiving any response traffic
+	sendSeqSent    uint        // Incremented each time we start sending
+	sendSeqRecv    uint        // Incremented each time we finish sending
 }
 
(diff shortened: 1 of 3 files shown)


neilalexander:
The latest commits seem to be better, I’ve been using VNC today over a couple of nodes with aggregate links running this build and it has been mostly OK. Still a small number of out-of-order packets and retransmissions but certainly less than on `develop` and not enough to create any real issues in my light testing.

Issue #60, opened 2018-03-12
Unit tests


Arceliar:
Go has nice machinery for writing unit tests. We should probably be using it. The sim can already be run manually to sort-of test some things, but it would make sense to unit test some parts (encoding/decoding packets, address<->nodeID conversion, etc). We could maybe rewrite a part of the sim to use the normal go testing framework, and then run it on a small network hard-coded network as a vertical slice test.

Arceliar:
Thinking aloud, a few basic kinds of tests we should work on:

1. Data formats. Make sure wire packets encode/decode as expected. Make sure decoders respond sanely when given garbage. It makes the most sense to do these after #42 is handled.
2. Crypto. Encrypted packets should decrypt. Garbage shouldn't decrypt. Encrypted packets with the wrong key should signal that there was a problem instead of claiming success (and returning garbage). Double check that this applies for nonces. Make sure nonces progress the expected way (they're used like sequence numbers).
3. Session logic. Make sure sessions are added and removed when they're supposed to (check that a timed out session gets removed when adding a new one, for example). Check that the window of old nonces works (prevent duplicate packets from the same nonce from getting through, but still allow old nonces to get through if they're only slightly out-of-order).
4. Switch / tree logic. Make sure we update roots/parents/coords when expected. Make sure lookups return the expected results.
5. Peer logic. Make sure we get coord updates when we're supposed to. Make sure it ignores updates when they're bad or if the peer has been throttled due to sending back information recently. I guess also check the signature stuff.
6. Network I/O tests?
7. Simulated networks. Make sure that things converge and everyone can ping everyone on a few simple topologies. We can take code from the sim for this. I'm thinking a simple 2 node test and then small networks with a ring, star, or grid topology. Note that some of these may take a while to converge. It would be good to make the network converge faster in general, so maybe we should work on that before adding the larger tests.

Arceliar:
We can't realistically create enough tests before v0.3 to close this issue, so moving to v0.4, but we should make it a priority.

PR #562, opened 2019-09-28
#60 Add simple tests and benchmark

AwesomePatrol:
This pull request adds:
 * test for peering connection between two nodes (on localhost),
 * test for message transfer between two nodes in both directions,
 * benchmark for transfer between nodes.

Tests should be considered sanity checks as they will not direct the developer to the particular point of failure.
Benchmark makes it easy to pinpoint places where performance can be improved when run with `cpu`/`memprofiler` or similar tools.

### DIFF
--- a/src/yggdrasil/core_test.go
+++ b/src/yggdrasil/core_test.go
@@ -30,7 +30,7 @@ func GetLoggerWithPrefix(prefix string) *log.Logger {
 	return l
 }
 
-func CreateAndConnectTwo(t *testing.T) (*Core, *Core) {
+func CreateAndConnectTwo(t testing.TB) (*Core, *Core) {
 	nodeA := Core{}
 	_, err := nodeA.Start(GenerateConfig(), GetLoggerWithPrefix("A: "))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


AwesomePatrol:
The changes are simple, but very helpful as I did not find any tests outside `misc/`. 

Due to number of asynchronous events, `WaitConnected` is ugly. It is possible to hide this with higher level functions, like in [gomega](https://godoc.org/github.com/onsi/gomega#Eventually). Some developers like to keep their code "pure" and try to keep using only standard library. I would like to know if you are open to pulling dependencies for test frameworks.

Arceliar:
Looks sensible to me. As far as dependencies go, I think I'd be more willing to accept new dependencies for testing code than in the main code, since the tests won't find their way into the final binary, but I'd still generally prefer not to bring in dependencies for tests if there's not a good reason for it. If it dramatically simplifies the way a particular test is written, then it's worth considering. I suspect that at this stage, where we have basically no tests written, there are enough things that would benefit from small/simple unit tests that we *probably* don't need to bring any larger frameworks into the picture yet.

neilalexander:
I’m very much with @Arceliar on this - if there are ways we can run tests that don’t depend on external frameworks then ideally we should do that, but I won’t be too upset if we have some small dependencies for tests if they don’t make their way into regular non-test builds.

PR #563, opened 2019-10-01
Systemd: tun module and capabilities

Arano-kai:
- Enable (and limit to) capabilities that require to setup tun/tap interface.
- Ensure that tun module is active.


Arceliar:
This looks OK to me, but @neilalexander handles the packaging and init/config, so I'll wait to see what he thinks.

PR #567, opened 2019-10-05
Reduce latency under load

Arceliar:
EDIT: needs more work, the bufferbloat is better but throughput suffers with high latency links.

In my tests, this changeset drastically reduces latency when under load. The parts that matter are the 1 line change in tcp.go and the change to writeMsg(s) in stream.go, the rest is just updated to match the slight change in API needed by the writeMsg change.

Between network namespaces with 10ms latency and 10 Mbps bandwidth limits attached to the link (so 20 ms round trip), when running iperf in the background to keep the link saturated, the latency before this patch would consistently climb to around the 2000 ms range and hover there. After this patch, the latency varies between around 40 and 120 ms in the best case, depending on MTU, and around 80 to 180 MTU in the worst case (less stable for small packets -- large packets seem to slowly climb in latency and then drop back down again).

Unfortunately, this does reduce speed someone for small packets, since it no longer bunches up many small packets to write them together in 1 syscall. There's room to improve this in the future, by delaying when the buffer flushes until the link is idle, but I'm not completely certain how to make that play nice with the timers to detect blocks on syscalls, so that's something to investigate in the future. Since sending many small packets is slower either way, I think cutting the latency by a factor of ~50 is worth more than the bandwidth we lose from this (it drops from ~600-700 to ~400-500 on my machine).

I also updated to the latest version of my `phony` library for the actor model stuff, though I expect that not to have noticeable affects on performance for this application.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,3 +17,5 @@ require (
 	golang.org/x/sys v0.0.0-20190813064441-fde4db37ae7a
 	golang.org/x/text v0.3.2
 )
+
+go 1.13
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/yggdrasil-network/yggdrasil-go
 
 require (
-	github.com/Arceliar/phony v0.0.0-20190907031509-af5bdbeecab6
+	github.com/Arceliar/phony v0.0.0-20191004004458-c7ba8368bafa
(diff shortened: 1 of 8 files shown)


Arceliar:
Closing until I fix (or at least better understand) some issues.

PR #569, opened 2019-10-05
Enable bbr for tcp sockets on linux

Arceliar:
Using the bbr congestion control mechanism on linux can massively reduce latency when under load. Ideally, this should be set system-wide with `sysctl -w net.ipv4.tcp_congestion_control=bbr`, but that's seems invasive. This patch uses a sockopt to set bbr congestion control on links to ygg peers.

You still get better performance if the endpoints of a connection have bbr enabled, since then they also react to the full end-to-end bufferbloat (including ygg's internal buffers), but enabling it at the link level seems to give *most* of the benefits (latency is ~25-33% as large as without enabling bbr).

I'm debating whether or not it makes sense to also *optionally* set it system-wide, but set that option to enabled by default, so linux users (with a recent enough kernel) have things "just work". I hesitate because that's kind of messy, since it would affect *all* applications running on the system, and it would remain in effect even after ygg exits.

### DIFF
--- a/src/yggdrasil/tcp_linux.go
+++ b/src/yggdrasil/tcp_linux.go
@@ -19,10 +19,14 @@ func (t *tcp) tcpContext(network, address string, c syscall.RawConn) error {
 		bbr = unix.SetsockoptString(int(fd), unix.IPPROTO_TCP, unix.TCP_CONGESTION, "bbr")
 	})
 
-	switch {
-	case bbr != nil:
-		return bbr
-	default:
-		return control
+	// Log any errors
+	if bbr != nil {
+		t.link.core.log.Debugln("Failed to set tcp_congestion_control to bbr for socket, SetsockoptString error:", bbr)
 	}
+	if control != nil {
+		t.link.core.log.Debugln("Failed to set tcp_congestion_control to bbr for socket, Control error:", control)
+	}
+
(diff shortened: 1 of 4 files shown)


src/yggdrasil/tcp_linux.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// WARNING: This context is used both by net.Dialer and net.Listen in tcp.go
+
+func (t *tcp) tcpContext(network, address string, c syscall.RawConn) error {
+	var control error
+	var bbr error
+
+	control = c.Control(func(fd uintptr) {
+		// sys/socket.h: #define	SO_RECV_ANYIF	0x1104
```
neilalexander: Not a problem with the code per-se but this comment is from the macOS multicast options.

Arceliar:
> I am happy with setting this on the socket itself but I’m very cautious about Yggdrasil enabling this systemwide even optionally, especially since many users may not understand the implications and we have no means to clean it up after a crash. On a multiuser system someone may configure one thing in `yggdrasil.conf` and another thing in `sysctl.conf`.

Agreed, lets take this patch as it is and then update the linux platform page documentation to suggest setting bbr in `sysctl.conf` as a possible OS-level optimization.

PR #578, opened 2019-10-19
Have listen and dial return a net.Conn

Arceliar:
This PR adjust the `Listener` to return a `(net.Conn, err)` so as to match that part of the `net.Listener` interface. In doing so, the interface of `yggdrasil.Conn` was changed slightly to return `net.Addr` from `LocalAddr` and `RemoteAddr`, and the tun/tap code was updated to type assert things as needed to work with this interface.

Marked as unfinished / work-in-progress since I'm sure there's probably more low-hanging fruit we could add to this (e.g. match the `Dialer` interface too).

### DIFF
--- a/src/address/address.go
+++ b/src/address/address.go
@@ -2,7 +2,11 @@
 // Of particular importance are the functions used to derive addresses or subnets from a NodeID, or to get the NodeID and bitmask of the bits visible from an address, which is needed for DHT searches.
 package address
 
-import "github.com/yggdrasil-network/yggdrasil-go/src/crypto"
+import (
+	"fmt"
+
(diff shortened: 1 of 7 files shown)


Arceliar:
This should now both match the `net.Listener` interface and the (undefined) interface satisfied by a `net.Dialer` struct, with `yggdrasil.Conn` satisfying the `net.Conn` interface (and being returned as such) from the provided functions.

Note that the `yggdrasil.Conn` is still unreliable and datagram-oriented, like a `net.UDPConn`, not reliable and stream oriented the way a `net.TCPConn` is, so users should be aware of that (it's not a drop-in replacement for any `net.Conn` that requires stream semantics, such as http, though some kind of reliable stream-oriented wrapper could be added in the future).

PR #583, opened 2019-10-23
Define module.Module interface

### LINKED ISSUE #581: AdminListen: none but yggdrasil opens the 9001 port
yggdrasil.conf:
```
...
# Listen address for admin connections. Default is to listen for local
# connections either on TCP/9001 or a UNIX socket depending on your
# platform. Use this value for yggdrasilctl -endpoint=X. To disable
# the admin socket, use the value "none" instead.
AdminListen: none
...
```

```
# netstat -natup |grep yggdrasil
...
udp6       0      0 :::9001             :::*        4167/yggdrasil
```

neilalexander:
This PR creates a `module.Module` interface which defines the functions that an Yggdrasil module must implement.

It also updates the `tuntap`, `multicast` and `admin` modules to comply with this interface.

This also fixes #581 as the multicast reload behaviour has been improved in this PR.


src/tuntap/tun.go
```diff
@@ -177,6 +180,11 @@ func (tun *TunAdapter) _start() error {
 	return nil
 }
 
+// IsStarted returns true if the module has been started.
+func (tun *TunAdapter) IsStarted() bool {
+	return tun.isOpen
```
Arceliar: Accessing `tun.isOpen` isn't thread safe, and probably needs to be in a `phony.Block` to run in the actor.

src/multicast/multicast.go
```diff
@@ -80,8 +86,14 @@ func (m *Multicast) Start() error {
 	return nil
 }
 
+// IsStarted returns true if the module has been started.
+func (m *Multicast) IsStarted() bool {
+	return m.isOpen
```
Arceliar: I just noticed this when `isStarted` was added, but I think even the existing use of `isOpen` (and possible some other multicast functions) may not be thread safe. This function certainly is not, since `IsStarted` is both usable outside the library and called within the actor in some of the code below. We probably need the exported `IsStarted` code to use a `phony.Block` to check this flag, and then skip the function and check the flag directly when we need it from inside the actor.

PR #588, opened 2019-10-23
Initial connection upgrade/TLS steganography 

neilalexander:
This is an experiment to implement peerings over TLS connections. 

This PR does nothing to Yggdrasil's own encryption, so a peering over TLS actually is encrypted twice. This might have performance penalties.

One thing to be aware of is that there's *absolutely no peer validation whatsoever* at the TLS layer—no certificate validation, no hostname verification, nothing. It's good for steganography (e.g. hiding the fact that it's an Yggdrasil connection) but not much else at this stage.

It works by implementing `TcpUpgrade`, which is a descriptor containing a new protocol name and a function for *replacing* the existing `net.Conn` with a new one that implements other security layers—in this case, `tls.go` replaces the `net.Conn` by wrapping it in a `tls.Conn`.

If no `TcpUpgrade` is specified, a regular peering is made exactly as they are today.

You can test this by specifying `tls://` instead of `tcp://` for both peers and listeners.

----

Benchmarks are as follows:

Link MTU 9000, Yggdrasil MTU 65535, with TLS:
```
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-10.00  sec  2.00 GBytes  1.71 Gbits/sec                  sender
[  4]   0.00-10.00  sec  1.99 GBytes  1.71 Gbits/sec                  receiver
```

Link MTU 9000, Yggdrasil MTU 65535, without TLS: 
```
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-10.00  sec  2.50 GBytes  2.15 Gbits/sec                  sender
[  4]   0.00-10.00  sec  2.50 GBytes  2.15 Gbits/sec                  receiver
```

Link MTU 9000, Yggdrasil MTU 1280, with TLS:
```
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-10.00  sec   202 MBytes   169 Mbits/sec                  sender
[  4]   0.00-10.00  sec   200 MBytes   168 Mbits/sec                  receiver
```

Link MTU 9000, Yggdrasil MTU 1280, without TLS:
```
[ ID] Interval           Transfer     Bandwidth
[  4]   0.00-10.00  sec   319 MBytes   267 Mbits/sec                  sender
[  4]   0.00-10.00  sec   318 MBytes   266 Mbits/sec                  receiver
```

### DIFF
--- a/src/yggdrasil/tls.go
+++ b/src/yggdrasil/tls.go
@@ -38,6 +38,7 @@ func (t *tcptls) init(tcp *tcp) {
 
 	certBuf := &bytes.Buffer{}
 
+	// TODO: because NotAfter is finite, we should add some mechanism to regenerate the certificate and restart the listeners periodically for nodes with very high uptimes. Perhaps regenerate certs and restart listeners every few months or so.
 	pubtemp := x509.Certificate{
 		SerialNumber: big.NewInt(1),
 		Subject: pkix.Name{
@@ -65,7 +66,7 @@ func (t *tcptls) init(tcp *tcp) {
 	t.config = &tls.Config{
 		RootCAs: cpool,
 		Certificates: []tls.Certificate{
-			tls.Certificate{
+			{
 				Certificate: [][]byte{derbytes},
 				PrivateKey:  edpriv,
 			},
--- a/src/yggdrasil/tls.go
+++ b/src/yggdrasil/tls.go
@@ -1,0 +1,92 @@
+package yggdrasil
+
+import (
+	"bytes"
+	"crypto/ed25519"
+	"crypto/rand"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/hex"
+	"encoding/pem"
+	"log"
+	"math/big"
+	"net"
+	"time"
+)
+
+type tcptls struct {
+	tcp         *tcp
+	config      *tls.Config
+	forDialer   *TcpUpgrade
+	forListener *TcpUpgrade
+}
+
+func (t *tcptls) init(tcp *tcp) {
+	t.tcp = tcp
+	t.forDialer = &TcpUpgrade{
+		upgrade: t.upgradeDialer,
+		name:    "tls",
+	}
+	t.forListener = &TcpUpgrade{
+		upgrade: t.upgradeListener,
+		name:    "tls",
+	}
+
+	edpriv := make(ed25519.PrivateKey, ed25519.PrivateKeySize)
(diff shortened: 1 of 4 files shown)


src/yggdrasil/tls.go
```diff
@@ -0,0 +1,92 @@
...
+
+	t.config = &tls.Config{
+		RootCAs: cpool,
+		Certificates: []tls.Certificate{
+			tls.Certificate{
+				Certificate: [][]byte{derbytes},
+				PrivateKey:  edpriv,
+			},
+		},
+		InsecureSkipVerify: true,
```
neilalexander: Ordinarily this would be criminal, but since we already have our own encryption going on anyway (even within a TLS session) and we have no other means to verify the trust of a TLS Yggdrasil node right now, I don't see a way around this.

If we disable this, then the TLS library will want to verify things like the connection hostname vs the certificate common name and so on, which then means that we have to place trust on an external source as to whether an Yggdrasil node is who they claim to be.

My suggestion at this stage is that, going forward, we look to implement public-key pinning for both `tcp://` and `tls://` so that you can be sure that no one has intercepted the peering connection.

Arceliar: Agreed, some kind of (optional) public key pinning should be added to the connection strings in the future.

src/yggdrasil/tls.go
```diff
@@ -0,0 +1,92 @@
...
+
+	certBuf := &bytes.Buffer{}
+
+	pubtemp := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			CommonName: hex.EncodeToString(tcp.link.core.sigPub[:]),
+		},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().Add(time.Hour * 24 * 365),
```
Arceliar: So I guess we need to restart our listeners at least once per year? We should probably document that so we remember to actually do so somewhere else in the code, eventually, in case things ever reach that level of stability. I imagine some of the remaining v0.3.5 nodes may have already been on for >6 months with no reboots, so this is plausibly something we need to think about.

src/yggdrasil/tls.go
Arceliar: Agreed, in our case it shouldn't matter since we already encrypt things internally. Either way, I 100% agree some kind of key pinning would be good to add to `tcp://` and `tls://` eventually.

Vort:
I heard that Telegram software implemented FakeTLS protocol.
Such approach may be faster than full TLS implementation, but less reliable in terms of DPI resilience.

PR #604, opened 2019-11-10
Add -address and -subnet command line options to cmd/yggdrasil

neilalexander:
This adds `-address` and `-subnet` command line arguments to `cmd/yggdrasil/main.go` which allow you to quickly get the IP address and/or subnet from your Yggdrasil config. This is a commonly requested feature.

Example usage:
```
./yggdrasil -useconffile /etc/yggdrasil.conf -address
201:6243:84d5:813a:2a98:e6a:4279:318
```
```
./yggdrasil -useconffile /etc/yggdrasil.conf -subnet
301:6243:84d5:813a::/64
```

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"bytes"
-	"crypto/sha512"
 	"encoding/hex"
 	"encoding/json"
 	"flag"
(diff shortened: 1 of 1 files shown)


wfleurant:
pull some trigger on this harmless feature, ci broken?

cmd/yggdrasil/main.go
```diff
@@ -188,6 +193,26 @@ func main() {
 	if cfg == nil {
 		return
 	}
+	// Have we been asked for the node address yet? If so, print it and then stop.
+	switch {
+	case *getaddr:
+		if pubkey, err := hex.DecodeString(cfg.EncryptionPublicKey); err == nil {
+			nodeid := sha512.Sum512(pubkey)
```
Arceliar: It would be better to use something like `crypto.GetNodeID(crypto.BoxPubKey(pubkey))` instead of importing and directly using sha512 here. The details of the key->ID->address mapping should be opaque, in case we ever need to change them for some reason (e.g. post-quantum encryption).

cmd/yggdrasil/main.go
```diff
@@ -188,6 +193,26 @@ func main() {
...
+			fromnodeid := address.AddrForNodeID((*crypto.NodeID)(&nodeid))
+			fmt.Println(net.IP(fromnodeid[:]).String())
+		}
+		return
+	case *getsnet:
+		if pubkey, err := hex.DecodeString(cfg.EncryptionPublicKey); err == nil {
+			nodeid := sha512.Sum512(pubkey)
```
Arceliar: Same as above

PR #605, opened 2019-11-11
Src version: return unknown not yggdrasilctl

wfleurant:


neilalexander:
This looks like an interesting typo that must have been introduced at some point, although I think in the past we returned `yggdrasil` from `GetBuildName` rather than `unknown`. Might be better to use that.

wfleurant:
it was discovered with some openwrt-yggrasil builds populating NodeInfo with yggdrasilctl and unknown. that bug has been fixed in the owrt Makefile.

Sent from ProtonMail mobile

-------- Original Message --------

neilalexander:
@Arceliar Do you think it is better in this instance to return `yggdrasil` or `unknown`?

PR #607, opened 2019-11-14
Systemd: move config generation to a separate unit

Arano-kai:
- Modular unit composition: different tasks in separate units
- Use systemd tool set to run checks
- Avoid using inline shell in unit


neilalexander:
Has this been tested on a range of systemd versions?

neilalexander:
I've assigned this to v0.3.13. Once v0.3.12 is released, we will merge this to `develop` to give people time to test.

PR #613, opened 2019-11-21
Add API functions for manipulating maximum session MTU

neilalexander:
The problem today is that the `tuntap` module would wrap the MTU down to a sane size but the session code in `Core` wouldn't. This would mean that setting a value larger than 65535 would cause a `uint16` overflow and then traffic wouldn't send over the session.

This adds new API functions for setting the maximum MTU, and then updates `tuntap` to call those as necessary, so that the session code is notified about the *real* TUN/TAP MTU.

For the sake of solving this problem for v0.3.12, I don't plan to make any other big changes, but it probably makes sense to fail hard or ensure that the `NodeConfig` expects a `uint16` rather than an `int` to make these kinds of mistakes more obvious. 

This also removes one of our awful shared-config-struct references, which I will shortly be on a mission to nuke all of them from `src/yggdrasil`.

### DIFF
--- a/src/yggdrasil/api.go
+++ b/src/yggdrasil/api.go
@@ -365,18 +365,19 @@ func (c *Core) SetNodeInfo(nodeinfo interface{}, nodeinfoprivacy bool) {
 
 // GetMaximumSessionMTU returns the maximum allowed session MTU size.
 func (c *Core) GetMaximumSessionMTU(mtu uint16) uint16 {
-	return c.router.sessions.myMaximumMTU
+	mtu := 0
+	phony.Block(c.router, func() {
+		mtu = c.router.sessions.myMaximumMTU
+	})
+	return mtu
 }
 
-// SetMaximumSessionMTU sets the maximum allowed session MTU size. The return
-// value contains the actual set value, since Yggdrasil will not accept MTUs
-// below 1280 bytes. The default value is 65535 bytes.
-func (c *Core) SetMaximumSessionMTU(mtu uint16) uint16 {
-	if mtu < 1280 {
-		mtu = 1280
-	}
-	c.router.sessions.myMaximumMTU = mtu
-	return mtu
+// SetMaximumSessionMTU sets the maximum allowed session MTU size. The default
+// value is 65535 bytes.
+func (c *Core) SetMaximumSessionMTU(mtu uint16) {
+	phony.Block(c.router, func() {
+		c.router.sessions.myMaximumMTU = mtu
+	})
 }
 
 // GetNodeInfo requests nodeinfo from a remote node, as specified by the public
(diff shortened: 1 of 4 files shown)


src/yggdrasil/api.go
```diff
@@ -363,6 +363,22 @@ func (c *Core) SetNodeInfo(nodeinfo interface{}, nodeinfoprivacy bool) {
 	c.router.nodeinfo.setNodeInfo(nodeinfo, nodeinfoprivacy)
 }
 
+// GetMaximumSessionMTU returns the maximum allowed session MTU size.
+func (c *Core) GetMaximumSessionMTU(mtu uint16) uint16 {
+	return c.router.sessions.myMaximumMTU
```
Arceliar: This field can only be safely read/written by the router actor (since it needs to check it to create sessions), unless we can guarantee that `c.router.sessions.myMaximumMTU` remains constant after it is first set (and I don't think we can guarantee that).

So this and `SetMaximumSessionMTU` below both need to go through the actor. A `phony.Block` would probably be easiest.

src/yggdrasil/api.go
```diff
@@ -363,6 +363,22 @@ func (c *Core) SetNodeInfo(nodeinfo interface{}, nodeinfoprivacy bool) {
...
+}
+
+// SetMaximumSessionMTU sets the maximum allowed session MTU size. The return
+// value contains the actual set value, since Yggdrasil will not accept MTUs
+// below 1280 bytes. The default value is 65535 bytes.
+func (c *Core) SetMaximumSessionMTU(mtu uint16) uint16 {
+	if mtu < 1280 {
+		mtu = 1280
+	}
+	c.router.sessions.myMaximumMTU = mtu
```
Arceliar: Like above, this needs to go through a `phony.Block` through the router to make it thread safe. Also, I'm not sure we should be enforcing the 1280 limit here, since that applies to IPv6 only. Maybe the tuntap should enforce that limit, and we just trust it here (and enforce the 65535 limit, but that's implicit from the `uint16` already).

I think the router also needs to send a message to each active session to have them update their local `sinfo.myMTU` value.

src/yggdrasil/session.go
```diff
@@ -187,9 +189,9 @@ func (ss *sessions) createSession(theirPermKey *crypto.BoxPubKey) *sessionInfo {
 	sinfo.mySesPriv = *priv
 	sinfo.myNonce = *crypto.NewBoxNonce()
 	sinfo.theirMTU = 1280
-	ss.router.core.config.Mutex.RLock()
-	sinfo.myMTU = uint16(ss.router.core.config.Current.IfMTU)
-	ss.router.core.config.Mutex.RUnlock()
+	// TODO: sinfo.myMTU becomes unnecessary if we always have a reference to the
```
Arceliar: I think `sinfo.myMTU` is still necessary for thread safety, unless we can guarantee that `ss.myMaximumMTU` is constant.

PR #619, opened 2019-11-25
Bugfix

Arceliar:
This attempts to prevent incorrect idle notification in the switch, which should (hopefully) help with some packet reordering issues.

It needs testing to make sure that other things don't break as a result, specifically that the flow of packets along a link doesn't become permanently blocked if a link is temporarily out and becomes "stalled" in the link code.


Vort:
I don't see any difference with `iperf3 -6 -c y.thingylabs.io -u` test.
Still lots of `OUT OF ORDER` messages.

Vort:
Wow! Looks like you have fixed 90% of the reordering problem. Thanks.
It still happens while connection is established, but after that problem is gone.

PR #630, opened 2019-11-30
Info log level for failed connect to peer

armatusmiles:
I think better to see in default log level when connect to peer is unsuccessful


armatusmiles:
> Since we currently try to re-add peers every minute or so, I am not sure if this will result in a lot of log spam.

You right. It is looks like every minute spam 
```
2019/12/01 13:35:04 Failed to dial TCP: dial tcp 51.37.137.221:37145: i/o timeout
2019/12/01 13:36:04 Failed to dial TCP: dial tcp 51.37.137.221:37145: i/o timeout
```
closed

PR #635, opened 2019-12-11
Add support for freebsd/arm64

dmgk:
http://thunderx1.nyi.freebsd.org/data/head-arm64-default/p518685_s355218/logs/errors/yggdrasil-0.3.11.log

@neilalexander Looks like you're also the FreeBSD port maintainer, could you please update the port too? (CC dmgk@freebsd.org)


neilalexander:
```
# golang.org/x/sys/unix
vendor/golang.org/x/sys/unix/ztypes_freebsd_arm64.go:400:12: undefined: uint128
```
Yikes, this looks like a problem in the Go `x/sys/unix` package, not in Yggdrasil itself. Which Go version are you building with?

dmgk:
go1.14beta1. See also https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=242312

neilalexander:
Since there is a commit that fixes the `uint128` thing, I have updated the `develop` branch with an updated `go.mod`/`go.sum` to point to the newer version of `golang.org/x/sys`. I wonder if this fixes your problem?

PR #636, opened 2019-12-11
util: fix possible OOB in IPv4 flowkey calc, use switch there

cathugger:
ihl may grow upto 15*4=60 so extract and check it before using it as offset in flowkey calculation.
also replace IFs with switches for protocol matching as it's less redundant and nicer to document.


neilalexander:
Thanks for this!

PR #641, opened 2020-01-04
Misc tuning

Arceliar:
Just a few relatively small optimizations that seemed like low hanging fruit.

1. `yggdrasilctl` now just calls `os.Exit` on some function, which returns an error code. That allows the internal function to safely use `defer` for cleanup etc, since this happens before the `os.Exit` call causes everything to shut down immediately.
2. In the DHT, always store the 2 closest nodes in each direction around the ring. That way, if your immediate neighbor disappears, you always already know the replacement. This *should* make things a little more robust under node joins/leaves, and *may* speed up bootstrapping a little.
3. The search code has been simplified significantly. Now, instead of storing a list of all visited nodes, we store the single closest node to the destination that we've gotten a response from. When a response comes in, we immediately send a request to *all* nodes in the response (in parallel) that are closer to the destination than the closest visited node. This makes each search take up constant space, and searches should finish more quickly (especially when a node fails to respond or when searching for an unreachable destination).

Not very well tested yet.

### DIFF
--- a/cmd/yggdrasilctl/main.go
+++ b/cmd/yggdrasilctl/main.go
@@ -25,14 +25,20 @@ import (
 type admin_info map[string]interface{}
 
 func main() {
+	// makes sure we can use defer and still return an error code to the OS
+	os.Exit(run())
+}
+
+func run() int {
 	logbuffer := &bytes.Buffer{}
 	logger := log.New(logbuffer, "", log.Flags())
-	defer func() {
+	defer func() int {
 		if r := recover(); r != nil {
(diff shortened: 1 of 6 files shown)


neilalexander:
This seems to be working quite well so far, although without a sim it's difficult to know if it's truly converging fully. Searches seem to be OK though in regular usage.

I brought the PR up-to-date with `develop` as that now includes #645 — the combination of the two is now running on the API server, but the API server shows quite a few more timeouts on DHT pings. I don't know if this is any worse than it was before though.

PR #643, opened 2020-01-05
Unify MTU datatypes across the codebase

### LINKED ISSUE #616: Make sure MTUs are using a common datatype throughout
The codebase bounces between `int` and `uint16` for MTU values. We should make these consistent and possibly type-alias something to make changing it in future easier, if possible.

We also have some code that tries to correct the MTU if a stupid value is given, but we may ultimately want to remove this and fail hard instead.

adamruzicka:
The codebase uses int and unit16 to represent MTU randomly. This change
unifies it to a MTU type from types package, which is currently uint16.

Partially fixes #616, didn't want to mix this with MTU correction.

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -22,7 +22,10 @@ import (
 
 	"github.com/yggdrasil-network/yggdrasil-go/src/crypto"
 	"github.com/yggdrasil-network/yggdrasil-go/src/defaults"
+	"github.com/yggdrasil-network/yggdrasil-go/src/types"
 )
+
+type MTU = types.MTU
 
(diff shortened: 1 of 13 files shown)


neilalexander:
Thanks for the PR! Two notes:

1. Please make PRs against the `develop` branch instead of `master` - should be able to edit the PR and merge forward without closing and reopening
1. It looks like CI has failed as some platform-specific code has not been updated

adamruzicka:
Hopefully the platform-specific stuff should be fixed, will try to change the PR base

adamruzicka:
And we're green, sorry about the mess

PR #647, opened 2020-01-31
Fixed initrc

rany0:
* Made routing work properly (sometimes yggdrasil does but due to some bug it doesn't add route quick enough) 
* Removed tap0 if it existed during start (and recreated it)
* Use the `/usr/local/etc/` instead of `/etc/` to comply with the FreeBSD hier. In FreeBSD, `/etc/` is reserved for the base system; `/usr/local/etc/` is meant for the third-party applications.


neilalexander:
I don't actually think the files in `contrib` are used anymore (we maybe should remove them altogether). The FreeBSD port is based on the https://github.com/yggdrasil-network/yggdrasil-package-freebsd/ repo so might be best to make fixes there!

PR #651, opened 2020-02-08
Search

Arceliar:
More search optimizations and debug level logging.

Each search response now starts up a separate thread to continue the search. These threads end as soon as they no longer contain any nodes to attempt that are closer than the closest node (in keyspace) to respond so far. This is meant to help prevent various DoS attacks where a malicious node would block the search by returning a large number of results with bad coords, to try to block the search from progressing (which now only blocks the malicious node's thread of the search).

### DIFF
--- a/src/yggdrasil/conn.go
+++ b/src/yggdrasil/conn.go
@@ -130,8 +130,8 @@ func (c *Conn) search() error {
 					close(done)
 				}
 			}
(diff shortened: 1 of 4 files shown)


Vort:
Not sure how to test it, but node works at least.
20h of uptime and no visible problems.

PR #652, opened 2020-02-16
Use public keys in API functions

neilalexander:
This adds support for dialling with `"curve25519"` instead of `"nodeid"`, and also updates `RemoteAddr`, `LocalAddr` etc to return public keys instead of node IDs.

### DIFF
--- a/src/crypto/crypto.go
+++ b/src/crypto/crypto.go
@@ -199,9 +199,9 @@ func (k BoxPubKey) String() string {
 	return hex.EncodeToString(k[:])
 }
 
-// Network returns "pubkey" for "box" keys.
+// Network returns "curve25519" for "box" keys.
 func (n BoxPubKey) Network() string {
-	return "pubkey"
+	return "curve25519"
 }
 
 // NewBoxKeys generates a new pair of public/private crypto box keys.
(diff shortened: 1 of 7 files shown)


src/crypto/crypto.go
```diff
@@ -194,6 +194,16 @@ type BoxSharedKey [BoxSharedKeyLen]byte
...
 type BoxNonce [BoxNonceLen]byte
 
+// String returns a string representation of the "box" key.
+func (k BoxPubKey) String() string {
+	return hex.EncodeToString(k[:])
+}
+
+// Network returns "pubkey" for "box" keys.
+func (n BoxPubKey) Network() string {
+	return "pubkey"
```
Arceliar: We may eventually want to change this to `"curve25519"` or something, for if/when we have multiple key types.

src/yggdrasil/conn.go
```diff
@@ -350,16 +350,19 @@ func (c *Conn) Close() (err error) {
 	return
 }
 
-// LocalAddr returns the complete node ID of the local side of the connection.
-// This is always going to return your own node's node ID.
+// LocalAddr returns the complete public key of the local side of the
+// connection. This is always going to return your own node's node ID.
```
Arceliar: Should be "your own node's *public key*" in the last sentence.

PR #660, opened 2020-03-04
add apparmor support to debian package

rany0:


rany0:
This is kind of more urgent than AppArmor because `/etc/yggdrasil.conf` lets everyone see the private keys and other configuration which is very bad.

neilalexander:
What happens in this case if you install onto a system without AppArmor?

rany0:
It just adds the profile to `/etc/apparmor.d/` but does nothing. The script checks if AppArmor is available or not.

neilalexander:
Closing as I'm not sure I understand enough of what the consequences of this are and my ability to support the `.deb` is limited enough.

PR #665, opened 2020-03-19
More reliable search

Arceliar:
EDIT: original patch now in my search-serial branch for backup. This approach keeps the overall behavior closer to what's in master, while still trying to fix the same issues with e.g. getting stuck on dead ends if it comes across an outdated kad node.

In master, we keep track of the closest node visited so far in the search. Nodes which are not closer than this node are ignored.

With the patch, we keep track of a list of up to (currently) the 16 closest nodes visited so far. We then use the last node on the list as the threshold for deciding if we should visit somewhere -- unless the list isn't full yet, in which case we try anything. Either way, we ignore nodes if they're already on the list. There may be room to improve the implementation (reduce the number of sorts, keep a map to check in constant time if a node is already in the list, etc), just testing for now if things work. The point here is to give searches a chance to recover if they hit a dead-end somewhere by getting a response from a node that doesn't know everyone they should, but not be so fault-tolerant that we end up crawling the entire network if the destination happens to be offline.

Also, in the dht, we cache the info about every node that responds to a search lookup. We don't consider these important, so they time out fairly quickly, but it means we can reuse that info if we start another search. That potentially saves a couple of network hops if e.g. we follow a hyperlink from one website to another, or a website requests resources hosted elsewhere. This subjectively seems to make browsing between sites a little more responsive, but I'm not sure if it's a real effect or just my imagination.

### DIFF
--- a/src/yggdrasil/search.go
+++ b/src/yggdrasil/search.go
@@ -4,15 +4,13 @@ package yggdrasil
 
 // The basic idea is as follows:
 //  We may know a NodeID (with a mask) and want to connect
-//  We begin a search by initializing a list of all nodes in our DHT, sorted by closest to the destination
-//  We then iteratively ping nodes from the search, marking each pinged node as visited
-//  We add any unvisited nodes from ping responses to the search, truncating to some maximum search size
-//  This stops when we either run out of nodes to ping (we hit a dead end where we can't make progress without going back), or we reach the destination
-//  A new search packet is sent immediately after receiving a response
-//  A new search packet is sent periodically, once per second, in case a packet was dropped (this slowly causes the search to become parallel if the search doesn't timeout but also doesn't finish within 1 second for whatever reason)
-
-// TODO?
-//  Some kind of max search steps, in case the node is offline, so we don't crawl through too much of the network looking for a destination that isn't there?
+//  We beign a search by sending a dht lookup to ourself
+//  Each time a node responds, we sort the results and filter to only include useful nodes
+//  We then periodically send a packet to the first node from the list (after re-filtering)
+//  This happens in parallel for each node that replies
+//  Meanwhile, we keep a list of the (up to) 16 closest nodes to the destination that we've visited
+//  We only consider an unvisited node useful if either the list isn't full or the unvisited node is closer to the destination than the furthest node on the list
+//  That gives the search some chance to recover if it hits a dead end where a node doesn't know everyone it should
 
 import (
 	"errors"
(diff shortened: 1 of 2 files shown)


src/yggdrasil/search.go
```diff
@@ -185,13 +228,18 @@ func (s *searches) newIterSearch(dest *crypto.NodeID, mask *crypto.NodeID, callb
...
 	from := dhtInfo{key: res.Key, coords: res.Coords}
-	if *from.getNodeID() != sinfo.visited && dht_ordered(&sinfo.dest, from.getNodeID(), &sinfo.visited) {
-		// Closer to the destination, so update visited
-		sinfo.searches.router.core.log.Debugln("Updating search:", &sinfo.dest, from.getNodeID(), sinfo.send, sinfo.recv)
-		sinfo.visited = *from.getNodeID()
-		sinfo.time = time.Now()
-	}
 	them := from.getNodeID()
+	sinfo.closest = append(sinfo.closest, them)
+	sort.SliceStable(sinfo.closest, func(i, j int) bool {
```
neilalexander: Go’s `container/heap` package may provide some optimisation here rather than repeating sorts.

Arceliar: Yes, that could be a good option. I changed the approach I'm using in this PR, so it's not relevant to this spot in the code anymore, but maybe in other places where things are sorted.

src/yggdrasil/search.go
```diff
@@ -24,7 +24,8 @@ import (
 
 // This defines the time after which we time out a search (so it can restart).
 const search_RETRY_TIME = 3 * time.Second
-const search_STEP_TIME = 100 * time.Millisecond
+const search_STEP_TIME = time.Second
```
Arceliar: I should mention, this time is really arbitrary. I'm trying to error on the side of minimizing traffic by not sending packets too aggressively when things aren't timing out or reaching a dead end.

src/yggdrasil/search.go
```diff
@@ -4,15 +4,13 @@ package yggdrasil
...
-//  We begin a search by initializing a list of all nodes in our DHT, sorted by closest to the destination
-//  We then iteratively ping nodes from the search, marking each pinged node as visited
-//  We add any unvisited nodes from ping responses to the search, truncating to some maximum search size
-//  This stops when we either run out of nodes to ping (we hit a dead end where we can't make progress without going back), or we reach the destination
-//  A new search packet is sent immediately after receiving a response
-//  A new search packet is sent periodically, once per second, in case a packet was dropped (this slowly causes the search to become parallel if the search doesn't timeout but also doesn't finish within 1 second for whatever reason)
-
-// TODO?
-//  Some kind of max search steps, in case the node is offline, so we don't crawl through too much of the network looking for a destination that isn't there?
+//  We beign a search by sending a dht lookup to ourself
```
rex4539: beign -> begin

Arceliar: Good catch, thanks!

PR #668, opened 2020-03-26
Add DHTCrawler module

zhoreeq:
This module adds "getNetworkMap" Admin API command which returns the same output as yggcrawl.

I've just copy pasted parts from yggcrawl, so the code might be horrible. But it works! :)


Arceliar:
I have to veto this one. I don't really like that it's possible to crawl the DHT in the first place, and getting rid of it without hurting reliability is something I'm actively looking into for v0.4. Every time someone crawls the network, it costs bandwidth for every node (potentially a *lot* of bandwidth for nodes that need to route the DHT lookups). Crawling isn't a viable long-term strategy for anything, and if the project continues to grow then at some point we'll need to shut down the crawlers and retire the map and nodeinfo databases (crawling will take too long and cost too much bandwidth to be worth it). The last thing we should do is add that function to the core program, where people running yggdrasilctl could run it without knowing what it does or understanding the cost to the network.

Issue #658, opened 2020-02-24
v0.4 switch changes needed


Arceliar:
Sort-of a followup to #555.

Part way through v0.3, we added a set of queues to the switch. All packets are passed to the switch worker, which either forwards them immediately to an idle peer, or buffers them until a peer becomes idle. Initially, it would forward to *any* idle peer that is closer to the destination (in tree metric space), and buffer only if *all* of them are busy. However, this resulted in lots of packets arriving out-of-order, which higher layers of the network stack dislike (e.g. TCP throttles down to almost nothing).

Later on, we switched to sending to the "best" peer (minimum distance to the destination, with ties broken by minimum latency to the root) if that peer is idle, and otherwise buffering until a peer (any peer) becomes idle. This was done as a hotfix to make sure that (most) traffic tends to be forwarded to the "best" peer, since this prevents (most) packet reordering, which keeps the transport and application layers of the network stack much happier.

That hotfix needs to be removed in v0.4 and replaced with something better. There's basically 3 directions we could go:

1. Revert to forwarding across all paths, and then have some mechanism at the session level to reorder packets correctly (e.g. TCP-like retransmission or FEC). This adds a lot of complexity just to do the job that the higher layers of the stack are supposed to handle themselves, and even if we do this *perfectly* we'd end up killing performance for some applications (bandwidth would be great, but there'd be latency jitter every time a packet arrived out-of-order and needed to be buffered until the earlier packets arrive or are reconstructed from FEC). So this isn't really an option for now, as far as I'm concerned (it's too much work to implement something that isn't supposed to be our job in the first place, and it distracts from the part I actually care about -- compact routing).

2. Instead of a single set of queues in the switch, which all traffic goes through, keep a separate queue per peer. This would let us keep separate queues per traffic flow (so e.g. a bulk TCP transfer to one user doesn't block a small stream of UDP traffic from another user's VoIP call), so we'd get to keep the congestion control aspect of the current switch, we just wouldn't have the option to route around congested links (which we currently avoid anyway, because the higher layers of the stack hate it). The main benefit here is that it would simplify the switch logic enough that we can have each peer (actor) do switch lookups in parallel, so it removes a (theoretical) bottleneck from having a single central switch actor.

[...]

PR #675, opened 2020-04-05
Buffers

Arceliar:
Work-in-progress on the backwards compatible part of #658 (if we with source routing). Not intended for v0.3.X, just opening the PR to document work and make it available for testing. That said, I'm not entirely opposed to including it in v0.3.X if we do another release at some point.

Instead of all packets flowing through the switch, the switch prepares a lookup table and sends it to the router and the peer actors. The peer actors then use the lookup table to determine where to forward the packet. Each peer then maintains its own set of buffers for traffic flows. This allows switching to be somewhat more multi-threaded, as packets no longer pass through the (hypothetical) bottleneck of a single switch worker. With source routing, we would simply replace the use of the lookup table at each peer with a step where we read the next hop form the packet and update the packet's headers accordingly.

In addition, nonce handling at the session level has been modified to require increasing nonce with every packet received, since that's the behavior that we *eventually* want to strictly enforce at the session level. That will make more sense after future (backwards incompatible) changes to the session code and the way traffic flows are labeled for the switch/peer level buffering. For now, the net effect should be that packets always arrive in order or not at all.


Arceliar:
Merging this since I want other changes to be built on top of it.

PR #688, opened 2020-05-09
mDNS support

neilalexander:
This PR facilitates local node discovery through mDNS. Or rather, it will, when it's finished.


neilalexander:
This *sorta* works, aside from the fact that interface-specific listeners don't get set up properly right now (I am guessing something has changed elsewhere in ygg/ironwood but haven't looked *too* closely yet).

This may not respond to changes as quickly as the existing multicast code though, since we need to wait for the current lookup to finish (currently with a 5-second context) and re-scan before detecting record changes if another node restarts. The fact that it's query-driven also means that there's a lot of DNS queries sent out by multicast, even if no one else responds, so it ultimately feels like it'll be a lot more idle traffic on the local subnet than the existing multicast implementation.

Therefore not really sure if there's an obvious benefit to this.

Issue #314, opened 2019-01-18
Lint all the things


neilalexander:
We should clean up naming etc. `golint` has lots of suggestions for us.

neilalexander:
This was done in v0.4 and checked by CI.

PR #674, opened 2020-04-05
Fixed some linter issues

Klesomik:
I just fixed some issues from [golangci-lint](https://github.com/golangci/golangci-lint). There are many more issues but I decided to try at first with the most simple. Anyway, it is connected with issue #314.
Is it a good idea to use such a linter in CI? It can check a new code automatically before code review. Also, it is super fast and it checked all the project within a minute.

These are some issues which [golangci-lint](https://github.com/golangci/golangci-lint) found: 

`cmd/genkeys/main.go`
```go
cmd/genkeys/main.go:54:15: unslice: could simplify currentBest[:] to currentBest (gocritic)
        if isBetter(currentBest[:], newKey.id[:]) || len(currentBest) == 0 {
                    ^
cmd/genkeys/main.go:64:48: unslice: could simplify newKey.priv[:] to newKey.priv (gocritic)
                fmt.Println("sigPriv:", hex.EncodeToString(newKey.priv[:]))
                                                           ^
cmd/genkeys/main.go:65:47: unslice: could simplify newKey.pub[:] to newKey.pub (gocritic)
                fmt.Println("sigPub:", hex.EncodeToString(newKey.pub[:]))
                                                          ^
cmd/genkeys/main.go:66:47: unslice: could simplify newKey.id[:] to newKey.id (gocritic)
                fmt.Println("TreeID:", hex.EncodeToString(newKey.id[:]))
                                                          ^
cmd/genkeys/main.go:68:48: unslice: could simplify newKey.priv[:] to newKey.priv (gocritic)
                fmt.Println("boxPriv:", hex.EncodeToString(newKey.priv[:]))
                                                           ^
cmd/genkeys/main.go:69:47: unslice: could simplify newKey.pub[:] to newKey.pub (gocritic)
                fmt.Println("boxPub:", hex.EncodeToString(newKey.pub[:]))
                                                          ^
cmd/genkeys/main.go:70:47: unslice: could simplify newKey.id[:] to newKey.id (gocritic)
                fmt.Println("NodeID:", hex.EncodeToString(newKey.id[:]))
                                                          ^
```

`cmd/yggdrasil/main.go`
```go
cmd/yggdrasil/main.go:225:17: unslice: could simplify pubkey[:] to pubkey (gocritic)
            copy(box[:], pubkey[:])
                         ^
cmd/yggdrasil/main.go:63:5: S1004: should use bytes.Equal(conf[0:2], []byte{0xFF, 0xFE}) instead (gosimple)
[...]

PR #690, opened 2020-05-15
Accept some golint suggestions

yujiri8:
Partial work on #314. I'm new to this project, so I didn't tackle anything that involved adding comments or renaming things. Most of my changes are removing unnecessary `else`s when the `if` ends with a `return`, plus a couple of comment typos.


yujiri8:
Oh, actually I noticed Klesomik had already made PR #674, which partially overlapped with my changes, so I rebased to put mine on top of theirs.

Klesomik:
Great job! I think it worth adding golang-ci lint in CI and it will save a lot of time in the future.

Klesomik:
And maybe it is better to commit these changes not in master but in a different branch

neilalexander (approved):

neilalexander:
Thanks for this! I have also added `golangci-lint` into CI under #733 (although it doesn't fail on lint errors yet).

PR #692, opened 2020-05-16
Misc updates for the future branch

Arceliar:
EDIT: the other things I wanted to do here have been moved to #693, so I think the half of the changes that are here are OK on their own.

Updates to `future` to hopefully cover some edge cases where performance may have degraded compared to `develop`.

In particular: this should (hopefully) be faster about blocking and unblocking peers at the switch level, in cases where the TCP connection to a peer stops working. This still needs testing, so I'm not sure if it's done yet or not.

There's still work to do here, in particular I want to do something about switch queue sizes (currently they're hard coded, we should either read them from the config or find a reasonable way to make them zero configuration without hard coding a max size -- I think they mainly need to deal with bursts of traffic caused by e.g. TCP head-of-line blocking).

This PR also adjust the logic for how peers are inserted into the DHT, and how the DHT resets when a node's coords change. This should *hopefully* lead to faster DHT convergence after a node's coords reset (based on the sim results). It may slightly increase idle DHT traffic, by sending a DHT ping to peers immediately after receiving a switch/tree-level update from them. I expect this traffic to be small compared to the size of the (signed) update that triggers it.

The goal here is to get `future` into a state where it's possible to merge it into `develop` before we add any hard backwards compatibility breaks, just in case we need to add them to `v0.3.X` as part of a bugfix or something (I'm currently running `future` on my public node, the single switch worker in the stable builds was unable to handle the load caused by the current network, due to the way we queue everything together and then need to re-check distances whenever a link becomes idle).

### DIFF
--- a/src/yggdrasil/dht.go
+++ b/src/yggdrasil/dht.go
@@ -89,6 +89,11 @@ func (t *dht) reconfigure() {
 // Resets the DHT in response to coord changes.
 // This empties all info from the DHT and drops outstanding requests.
 func (t *dht) reset() {
+	for _, info := range t.table {
+		if t.isImportant(info) {
+			t.ping(info, nil)
+		}
+	}
 	t.reqs = make(map[dhtReqKey]time.Time)
 	t.table = make(map[crypto.NodeID]*dhtInfo)
 	t.imp = nil
@@ -144,12 +149,8 @@ func (t *dht) insert(info *dhtInfo) {
 
 // Insert a peer into the table if it hasn't been pinged lately, to keep peers from dropping
 func (t *dht) insertPeer(info *dhtInfo) {
-	oldInfo, isIn := t.table[*info.getNodeID()]
-	if !isIn || time.Since(oldInfo.recv) > dht_max_delay+30*time.Second {
-		// TODO? also check coords?
-		newInfo := *info // Insert a copy
-		t.insert(&newInfo)
-	}
+	t.insert(info)    // FIXME this resets timers / ping counts / etc, so it seems kind of dangerous
+	t.ping(info, nil) // This is a quick fix to the above, ping them immediately...
 }
 
 // Return true if first/second/third are (partially) ordered correctly.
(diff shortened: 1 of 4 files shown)


Arceliar:
So I think I have *most* of the queue stuff figured out on paper. Documenting it here in case I forget / can't get to it right away.

Currently, we just add packets to the queue as fast as we can, and then start dropping them when the queue gets too big. The problem is, we have no way to know if the queue is too big because the peer's TCP connection is blocked, or if the queue is too big because the runtime's scheduler hasn't gotten around to even trying to send the previous packet yet.

So the goal here is to detect when it's at least *possible* that the connection is blocked, and only drop packets under those circumstances. We drop 1 packet, then check if we're still blocked, and repeat as fast as we possibly can. In effect, this means our buffer sizes are "as small as possible" which I think is the right way to approach things.

Whenever the peer passes a packet to the link, it marks itself as non-idle. When a peer receives a packet (or has just sent a packet but has a non-empty queue), if it is not idle, then it sends a message to the link asking the link to respond if it's blocked. This message will contain a sequence number, which is incremented every time the peer sends a message.

When the link writer begins to write a packet to the socket, it sends a message to the link, letting it know that it's busy writing. As soon as it finishes the write, it sends another message to the link, letting it know that it's no longer busy.

When the link receives a message from the writer that says it's busy, it sets a bool flag to track this. When the link receives a message from the writer that says it's no longer busy, it unsets the flag. When the link receives a message from the peer asking if the link is blocked, it replies (just acks the sequence number back to the peer) if and only if it is currently busy (or maybe it acks either way, but includes the bool flag letting the peer know if it's busy or not).

When the peer receives a busy ack (or one with the flag set to true, if they're always responded to), it checks if the sequence number in the ack is the same as the link's current sequence number. If the acks do not match, then the message is ignored. If the acks match, then the peer drops 1 packet from the queue (according to whatever rules make sense), and if the queue is not empty, then it sends another message to the link asking if it's blocked. This will repeat until either the link unblocks and the peer is sent an idle notification (letting it send more packets), or until the peer has dropped all packets from the queue.

So the hard part here is the router/selfPeer interaction. There's no link struct involved, so currently this would mean we have to choose between buffering indefinitely or letting backpressure throttle us back. Buffering indefinitely is out of the question. Letting backpressure throttle us would slow down all traffic flowing *through* us (and could block the switch indefinitely if an application uses ygg as a library and handles connections incorrectly). So neither of those is an option. We'd probably need to add another actor between the router and the selfPeer to mediate this exchange, but I'm not yet sure what that would need to look like.

PR #702, opened 2020-05-25
Precompute more for the switch lookup table

Arceliar:
Work-in-progress.

I noticed that, at least on my public node, a significant fraction of its time is spent iterating over the map in the current lookup table and doing a dist calculation to find the next hop (the map iteration is actually the more expensive part, it seems, but distance calculations aren't too far behind).

This tries to get rid of the iteration by producing a nested series of maps. Lookups happen by looping over coords and lookup up that coord in the next layer of the nested map, stopping when we either get to the end or run into a dead end in the nested map structure.

I'm not 100% sure this always reproduces exactly the same next-hop selection as the current code in cases where there are multiple valid next hops. If it doesn't, then it's at least not obvious why.

### DIFF
--- a/src/yggdrasil/switch.go
+++ b/src/yggdrasil/switch.go
@@ -93,6 +93,20 @@ func (l *switchLocator) dist(dest []byte) int {
 	return dist
 }
 
+func (l *switchLocator) ldist(sl *switchLocator) int {
+	lca := -1
+	for idx := 0; idx < len(l.coords); idx++ {
+		if idx >= len(sl.coords) {
+			break
+		}
+		if l.coords[idx] != sl.coords[idx] {
+			break
+		}
+		lca = idx
+	}
+	return len(l.coords) + len(sl.coords) - 2*(lca+1)
+}
+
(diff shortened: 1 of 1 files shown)


src/yggdrasil/switch.go
```diff
@@ -510,87 +529,106 @@ func (t *switchTable) _handleMsg(msg *switchMsg, fromPort switchPort, reprocessi
...
+				here.next = make(map[switchPort]*tableElem)
+			}
+			var next *tableElem
+			var ok bool
+			if next, ok = here.next[elem.locator.coords[idx]]; !ok {
+				nextVal := *elem
+				next = &nextVal
+				here.next[next.locator.coords[idx]] = next
+			}
+			here = next
```
neilalexander, line 609: This isn’t run for the last iteration of the enclosing loop (`<` vs `<=`)?

Arceliar: Correct, this is setup for the next iteration of the loop. The last time through the loop, there's nothing to set up (`idx` would be out of range for `elem.locator.coords`), so this part is skipped.

PR #716, opened 2020-07-01
net.PacketConn

neilalexander:
This replaces the `net.Conn` API with `net.PacketConn`. 

Still to do:
- [ ] reinstate various address checks that were in `tuntap/conn.go`
- [ ] check actor behaviours
- [ ] fix `Resolve` blocking in TUN/TAP
- [ ] don't search for our direct peers - we already know their coords!
- [ ] close idle sessions
- [x] make sure we create a session if one doesn't exist
- [ ] finish implementing all of the `net.PacketConn` functions like deadlines etc


src/tuntap/iface.go
```diff
@@ -149,60 +148,43 @@ func (tun *TunAdapter) _handlePacket(recvd []byte, err error) {
...
-						return
-					}
-					for _, packet := range packets {
-						tc.writeFrom(nil, packet)
-					}
-				})
-				return
-			}()
+
+		_, boxPubKey, err = tun.core.Resolve(dstNodeID, dstNodeMask)
```
Arceliar, line 169: It looks like `Resolve` will block, so we should probably buffer this packet and start this operation in another goroutine (or later turn it into an actor thing to avoid the extra goroutine). I believe the old `Conn` code had something similar that you can steal, where it buffered things while waiting for a `Dial` to finish.

src/tuntap/tun.go
```diff
@@ -170,9 +166,11 @@ func (tun *TunAdapter) _start() error {
 	}
 	tun.core.SetMaximumSessionMTU(tun.MTU())
 	tun.isOpen = true
-	go tun.handler()
 	tun.reader.Act(nil, tun.reader._read) // Start the reader
 	tun.ckr.init(tun)
+	tun.packetConn.SetReadCallback(func(addr net.Addr, bs []byte) {
```
Arceliar, line 171: We should probably commit using the `PacketConn` interface at first, at least to test things, before we insert callbacks etc to work around it / make it actor-friendly. Or commit to using actors and don't worry about matching the `PacketConn` interface at all for now. Either way, mixing and matching is likely to hide bugs in one interface or the other.

src/yggdrasil/api.go
```diff
@@ -225,32 +226,47 @@ func (c *Core) GetSessions() []Session {
...
-// ConnDialers are stateless, you can request as many dialers as you like,
-// although ideally you should request only one and keep the reference to it for
-// as long as it is needed.
-func (c *Core) ConnDialer() (*Dialer, error) {
-	return &Dialer{
-		core: c,
-	}, nil
+// PacketConn returns a net.PacketConn which can be used to send and receive
+// information over Yggdrasil sessions.
+func (c *Core) PacketConn() *PacketConn {
```
Arceliar, line 231: So I think ideally we should make the `Core` export the `PacketConn` interface directly, to make it clear that there'e exactly 1 `PacketConn` available to the `Core`. As-is, people could be confused that multiple calls return a pointer to the same underlying `PacketConn` (which may break things outright if they try to use the `PacketConn` in existing code -- multiple things trying to read from the same underlying `PacketConn` could make them all think there's horrible packet loss, or they could call `Close` or the various `Deadline` functions and mess things up for eachother). At the very least, we should make it clear in the documentation that it's a pointer to the same `PacketConn`.

src/yggdrasil/packetconn.go
```diff
@@ -0,0 +1,277 @@
...
+		c.Act(nil, c._drainReadBuffer) // In case there's more
+	default:
+	}
+}
+
+// SetDeadline is equivalent to calling both SetReadDeadline and
+// SetWriteDeadline with the same value, configuring the maximum amount of time
+// that synchronous Read and Write operations can block for. If no deadline is
+// configured, Read and Write operations can potentially block indefinitely.
+func (c *PacketConn) SetDeadline(t time.Time) error {
```
Arceliar, line 209: As far as I can see, it looks like the `Deadline`s aren't currently implemented. We should probably document that for now. We probably need to think about how we can implement them and get the correct behavior. I can't think of a good way to do it (i.e. return in a bounded amount of time after the deadline without leaking any goroutines or something) without resorting to timers and `select`ing on channels. Which is OK for the blocking `PacketConn` interface I guess, and maybe an actor-based interface can just bypass it entirely with async message sends, but that seems a little ugly.

src/yggdrasil/packetconn.go
```diff
@@ -0,0 +1,277 @@
...
+func newPacketConn(ss *sessions) *PacketConn {
+	return &PacketConn{
+		sessions:   ss,
+		readBuffer: make(chan packet),
+	}
+}
+
+func (c *PacketConn) _sendToReader(addr net.Addr, payload []byte) {
+	if c.readCallback == nil {
+		c.readBuffer <- packet{
```
Arceliar, line 39: This should probably be in a `select` statement with a default case that drops the packet, otherwise forgetting to read would deadlock the whole node.

src/tuntap/iface.go
neilalexander: Yep, you are indeed right about that - it's on the to-do list, just wasn't sure of the best way to handle that.

src/tuntap/tun.go
neilalexander: Indeed - this was mostly just to satisfy the existing interface rather than breaking it, but it wouldn't be too difficult to change.

src/yggdrasil/api.go
neilalexander: Agreed on making `Core` implement `PacketConn`.

neilalexander:
This has been superseded by #800.

PR #717, opened 2020-07-03
Set default conf file on FreeBSD to /usr/local/etc/yggdrasil.conf

yujiri8:
FreeBSD generally uses /usr/local/ for everything installed in userspace.

May want to put this on hold until we figure out how to solve the main problem with the FreeBSD port (not being able to destroy the interface on shutdown).


yujiri8:
I believe this is ready now. If this gets merged, I can update the port Makefile to use the new version and Yggdrasil can work on FreeBSD without further issues.

neilalexander:
Sorry it took so long to merge this - thanks for submitting it.

PR #731, opened 2020-09-27
Version 0.3.15

neilalexander:
### Added
- Support for pinning remote public keys in peering strings has been added, e.g.
  - By signing public key: `tcp://host:port?ed25519=key`
  - By encryption public key: `tcp://host:port?curve25519=key`
  - By both: `tcp://host:port?ed25519=key&curve25519=key`
  - By multiple, in case of DNS round-robin or similar: `tcp://host:port?curve25519=key&curve25519=key&ed25519=key&ed25519=key`
- Some checks to prevent Yggdrasil-over-Yggdrasil peerings have been added
- Added support for SOCKS proxy authentication, e.g. `socks://user@password:host/...`

### Fixed
- Some bugs in the multicast code that could cause unnecessary CPU usage have been fixed
- A possible multicast deadlock on macOS when enumerating interfaces has been fixed
- A deadlock in the connection code has been fixed
- Updated HJSON dependency that caused some build problems

### Changed
- `DisconnectPeer` and `RemovePeer` have been separated and implemented properly now
- Less nodes are stored in the DHT now, reducing ambient network traffic and possible instability
- Default config file for FreeBSD is now at `/usr/local/etc/yggdrasil.conf` instead of `/etc/yggdrasil.conf`


Arceliar (approved):
I notice we're still on golang 1.14.1, while 1.15.2 is the latest stable release. But I'd rather not rock the boat at this point, we can think about updating go after the release is out. Looks good from this end.

PR #736, opened 2020-10-11
Fix AppArmor issue & have safe perms for config

rany2:
The AppArmor profile in contrib forbids `/usr/bin/yggdrasil` from reading the file in `/var/backups/yggdrasil.conf...`. This works around that restriction by having the shell do the reading of `/var/backups/yggdrasil.conf...` file while providing the same exact functionality without making the AppArmor profile less restrictive. 

Another change is the safe perms for the `/etc/yggdrasil.conf` (so that config will have 0640 permissions). This is important because if we kept the default of 644 then any user (privileged or unprivileged) will have the ability to read the yggdrasil private key. We use a restrictive umask of 0027 to make this possible.


analotia:
Regarding the AppArmor issues. This will not enable the AppArmor profile. All it does is provide fixes for the users of the AppArmor profile in contrib.

PR #737, opened 2020-10-14
Increase open file limit for systemd unit to 8192

zhoreeq:


neilalexander:
@Arceliar Did we decide what to do about DoS attacks by exhausting FDs?

neilalexander:
Closing as we should think of a better way to deal with this rather than bandaiding it. People may want to adjust their own FD limits too and IMO our package shouldn't interfere with that.

PR #738, opened 2020-10-18
Listener bugfix

Arceliar:
If the listener ever receives an error, then currently it will exit, and there's no mechanism to recover (other than maybe reconfiguring with a `SIGHUP`, I didn't test).

With this patch, instead of exiting when there's an error, check if the listener was stopped. If it wasn't stopped, then assume it's a temporary error, pause for 1 second (to avoid busy loops), and then continue trying to `Accept()` new connections.

Also, within the multicast code, currently we `panic` if we fail to get the list of network interfaces. Instead, return an empty map of interfaces/addresses if we failed to get them, as this can also be caused by temporary problems.

I'm not 100% happy with either of these changes, but they seem to (partially) work around a couple of issues reported in the chat. At the very least, if things enter into a broken state, they should now either recover on their own (eventually) or continue to spam the log with info about the problem.

### DIFF
--- a/src/multicast/multicast.go
+++ b/src/multicast/multicast.go
@@ -191,7 +191,9 @@ func (m *Multicast) getAllowedInterfaces() map[string]net.Interface {
 	// Ask the system for network interfaces
 	allifaces, err := net.Interfaces()
 	if err != nil {
-		panic(err)
+		// Don't panic, since this may be from e.g. too many open files (from too much connection spam)
+		// TODO? log something
+		return nil
 	}
 	// Work out which interfaces to announce on
 	for _, iface := range allifaces {
--- a/src/yggdrasil/tcp.go
+++ b/src/yggdrasil/tcp.go
@@ -226,7 +226,13 @@ func (t *tcp) listener(l *TcpListener, listenaddr string) {
 		sock, err := l.Listener.Accept()
 		if err != nil {
 			t.links.core.log.Errorln("Failed to accept connection:", err)
-			return
+			select {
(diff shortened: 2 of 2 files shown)


neilalexander (approved):
In principle looks fine, but would be keen to see it tested against exhausted file descriptors to be sure that just calling `Accept` again is sufficient and the listener isn’t just dead altogether after that error.

Arceliar:
> In principle looks fine, but would be keen to see it tested against exhausted file descriptors to be sure that just calling `Accept` again is sufficient and the listener isn’t just dead altogether after that error.

On my machine, it starts getting errors while the file descriptors are exhausted, but goes back to accepting connections after resources become available. We could probably tune the timeout before we give up on a new connection, to make them free up descriptors a little faster, but I'm not sure it's really that important (adjusting the timeout won't prevent an ongoing attack from blocking the listener, and the current timeout shouldn't run into problems except when under an attack, so there's little to be gained by messing with it... it would mostly just recover a few seconds faster if an attacker gives up).

neilalexander:
SGTM then!

zhoreeq:
What if an attacker will open more than 1024 connections per second from multiple hosts? It seems like the listener will be almost permanently in the "cooldown" mode, and it very unlikely will accept legitimate connections.

Vort:
@zhoreeq anti(D)DoS protection is a whole topic. Implementing it will not be easy. So I think it is better to make recovery after attack is stopped more smooth as the first step and after that think about more complicated measures.

PR #744, opened 2020-11-13
Fix DefaultIfName for OpenBSD

octeep:
Specifying the full path to the interface in OpenBSD would result in:
panic: Interface name must be tun[0-9]*

Therefore, DefaultIfName should be changed to tun0 in order to make yggdrasil work out of the box.


neilalexander:
Thanks for this!

Issue #437, opened 2019-06-17
Yggdrasil IP selected for public Internet


kravietz:
I'm experiencing a strange issue on a number of Linux boxes where the `tun0` interface IP is being selected as the source IP for outgoing connections even for non-Yggdrasil IPs:
```
# tcpdump -ni eth0 ip6
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
12:03:42.150824 IP6 201:77d2:16a9:6bc4:54fa:7fb5:e33f:a77f.46762 > 2001:41d0:701:1100::36e.39593: Flags [S], seq 1111678875, win 28800, options [mss 1440,sackOK,TS val 1852652957 ecr 0,nop,wscale 7], length 0
```
Obviously, this doesn't work because my node is trying to connect to a public node in order to join Yggdrasil network.

I suspect this might be caused by the fact that they are all being set with `medium` priority which results in the largest non-link-local subnet being selected (and that is `200::/7`):
```
# ip -6 route 
200::/7 dev tun0 proto kernel metric 256 pref medium
2a01:4f8:1c1c:7d3::/68 dev eth0 proto kernel metric 256 pref medium
fe80::/64 dev eth0 proto kernel metric 256 pref medium
fe80::/64 dev tun0 proto kernel metric 256 pref medium
default via fe80::1 dev eth0 metric 1024 pref medium
```

kravietz:
This seems to be more complicatd that just prefix-length. The problem can be essentially described as:
```
# ip route get 2001:41d0:701:1100::36e
2001:41d0:701:1100::36e from :: via fe80::1 dev eth0 src 201:77d2:16a9:6bc4:54fa:7fb5:e33f:a77f metric 1024 pref medium
```
So for connection to `2001::` address the kernel is selecting the `0201::` address on `tun0` rather than `2a01::` address on `eth0`.

After reading tons about `gai.conf`, [RFC6724](https://tools.ietf.org/html/rfc6724#section-5) and experimenting with `ip addrlabel` I haven't really made much progress apart from marking the Yggdrasil IP as "deprecated" (`preferred_lft 0`) thus preventing the kernel from selecting it on outgoing connections:
```
ip addr replace dev tun0 201:77d2:16a9:6bc4:54fa:7fb5:e33f:a77f/7 scope global preferred_lft 0
```
[...]

PR #748, opened 2020-12-06
Return ICMPv6 Destination Unreachable for unknown destinations

### LINKED ISSUE #180: can't ping if launch yggdrasil
I [haven't](url) ipv6 address from provider. I haven't any ipv6 address. But, if I launch yggdrassil, it's properly connected to the nodes and I can ping into yggdrasil network. But, I can't ping any clearnet ipv4 addresses. What I do wrong? How can I fix it?

neilalexander:
I have absolutely no idea if this will fix #180 and #437 but it's about the only thing that I can think of that might.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/vishvananda/netlink v1.0.0
 	github.com/vishvananda/netns v0.0.0-20190625233234-7109fa855b0f // indirect
+	github.com/yggdrasil-network/yggdrasil-extras v0.0.0-20201206140615-845aca5e7894 // indirect
 	golang.org/x/crypto v0.0.0-20200221231518-2aa609cf4a9d
 	golang.org/x/net v0.0.0-20200301022130-244492dfa37a
 	golang.org/x/sys v0.0.0-20200302150141-5c8b2ff67527
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,6 @@ require (
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/vishvananda/netlink v1.0.0
 	github.com/vishvananda/netns v0.0.0-20190625233234-7109fa855b0f // indirect
-	github.com/yggdrasil-network/yggdrasil-extras v0.0.0-20201206140615-845aca5e7894 // indirect
 	golang.org/x/crypto v0.0.0-20200221231518-2aa609cf4a9d
 	golang.org/x/net v0.0.0-20200301022130-244492dfa37a
 	golang.org/x/sys v0.0.0-20200302150141-5c8b2ff67527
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/vishvananda/netlink v1.0.0
 	github.com/vishvananda/netns v0.0.0-20190625233234-7109fa855b0f // indirect
+	github.com/yggdrasil-network/yggdrasil-extras v0.0.0-20201206140615-845aca5e7894 // indirect
```
Arceliar, line 15: Can you `go mod tidy` to see if this goes away? I see no reason why `yggdrasil-extras` should be included here (and it introduces a circular dependency), so I would expect this to not be needed.

go.sum
```diff
@@ -31,6 +31,8 @@ github.com/vishvananda/netlink v1.0.0 h1:bqNY2lgheFIu1meHUFSH3d7vG93AFyqg3oGbJCO
 github.com/vishvananda/netlink v1.0.0/go.mod h1:+SR5DhBJrl6ZM7CoCKvpw5BKroDKQ+PJqOg65H/2ktk=
 github.com/vishvananda/netns v0.0.0-20190625233234-7109fa855b0f h1:nBX3nTcmxEtHSERBJaIo1Qa26VwRaopnZmfDQUXsF4I=
 github.com/vishvananda/netns v0.0.0-20190625233234-7109fa855b0f/go.mod h1:ZjcWmFBXmLKZu9Nxj3WKYEafiSqer2rnvPr0en9UNpI=
+github.com/yggdrasil-network/yggdrasil-extras v0.0.0-20201206140615-845aca5e7894 h1:CiLvZWLT38MlyrfxhVVsnOqVBL/b5EUpXiOQnqNPLU0=
+github.com/yggdrasil-network/yggdrasil-extras v0.0.0-20201206140615-845aca5e7894/go.mod h1:xQdsh08Io6nV4WRnOVTe6gI8/2iTvfLDQ0CYa5aMt+I=
```
Arceliar, line 35: Same as the `go.mod` comment, I don't see any reason why this should be needed...

Arceliar (commented):
It looks like some unused dependencies may have found their way into `go.mod` and `go.sum`, which I'd like to have cleaned up before this is merged. I haven't tested it, but in general I think it makes sense to return an error such as this instead of silently dropping the packet. Whether or not applications or the OS will do the right thing is another matter... I could imagine some weird interactions may occur if CKR settings are modified after startup, so we may want to regression test this one a bit before merging it.

go.mod
neilalexander: I expect this has happened because I've done `./build -i` and it's been pulled in. `go mod tidy` has removed it.

Arceliar (approved):
Looks good to me, though I wouldn't be opposed to more testing (mostly changing CKR configuration at runtime) before merging into develop, just in case it does something unexpected.

neilalexander:
I've done some testing this evening and it seems that adding and removing CKR routes at runtime works fine with this—the ICMPv6 responses are sent when the CKR route is removed and they aren't when it is re-added.

Linux doesn't appear to cache the unreachability anywhere, presumably because it is a layer 3 tunnel and there's no Neighbour Discovery record in memory for the result to be cached in. Therefore pinging immediately after making changes seems to do the right thing.

Issue #732, opened 2020-09-27
Lint errors in CI should hard-fail


neilalexander:
We should hook up `golangci-lint` into CircleCI sometime.

neilalexander:
I think we did this, so closing.

PR #749, opened 2020-12-06
Tidy some lint errors

neilalexander:
This cleans up a bunch of lint errors by either cleaning up dead code, marking some things as `nolint` or making minor refactors.


neilalexander:
We can probably also do #732 too, once this is merged.

src/yggdrasil/core_test.go
```diff
@@ -1,3 +1,5 @@
+// +build debug
```
Arceliar, line 1: I'm not sure I like disabling this outside of debug builds, since it shouldn't be built outside of `go test` anyway, and hiding tests outside of debug builds may just lead to hiding errors (or problems with the tests)

src/yggdrasil/dialer.go
```diff
@@ -21,7 +21,7 @@ type Dialer struct {
 // "curve25519" or "nodeid" and the second parameter should contain a
 // hexadecimal representation of the target. It uses DialContext internally.
 func (d *Dialer) Dial(network, address string) (net.Conn, error) {
-	return d.DialContext(nil, network, address)
+	return d.DialContext(context.TODO(), network, address)
```
Arceliar, line 24: As mentioned in the yggdrasilsim code, we should remove the `ctx == nil` check elsewhere in the dial code, and use `context.Background()` here instead, I think.

src/yggdrasil/switch.go
```diff
@@ -474,10 +455,7 @@ func (t *switchTable) _handleMsg(msg *switchMsg, fromPort switchPort, reprocessi
 				return false
 			}
 		}
-		if sender.locator.root == t.core.sigPub {
-			return false
-		}
-		return true
+		return sender.locator.root == t.core.sigPub
```
Arceliar, line 458: It might just be that I haven't slept so I'm tired, but this looks like it returns the wrong value (should be `!=`), though I strongly dislike this change (it makes the logic less clear in a part of the code that's already hard to understand).

src/yggdrasil/wire.go
```diff
@@ -42,15 +42,6 @@ func wire_put_uint64(e uint64, out []byte) []byte {
...
 }
 
-// Returns the length of a wire encoded uint64 of this value.
-func wire_uint64_len(elem uint64) int {
-	l := 1
-	for e := elem >> 7; e > 0; e >>= 7 {
-		l++
-	}
-	return l
-}
```
Arceliar, line 52: I think we should keep this. We happen to not use it at the moment, but it will almost certainly be needed in the `future` when we source route.

Arceliar (commented):
I don't think we should work on linter errors right now. A couple of them unintentionally change behaviors in subtle ways that look like they're either broken (but it wouldn't show up outside of tests in dynamic networks), or may be broken on certain platforms. And a lot of the rest are changes to code that either isn't used now (but likely will be later), or code that's likely to be removed anyway because it will be replaced by something else (e.g. the Dial-related stuff). It's probably better to ignore the linter for now, since we know we're about to throw out a *lot* of code and rewrite a lot of what's left. It would be better to do this after the v0.4 changes are in and debugged.

neilalexander:
Will come back to this after conflicting changes are merged from `future`.

PR #751, opened 2020-12-19
Fix goroutine leak in link.go

Arceliar:
This should fix a bug in link.go where having a peer that continues to make connection attempts would leak goroutines until the original connection is closed.


neilalexander (approved):
Looks quite elegant!

PR #765, opened 2021-03-06
Upgrade to Go 1.16, update dependencies

neilalexander:
This moves our baseline up to Go 1.16 and updates all of our dependencies.


neilalexander:
CI error related to linting.

Arceliar (approved):
We should probably let it sit in develop for a little while before release, just in case any of the dependency updates causes unexpected changes to how something behaves.

PR #766, opened 2021-03-07
Build MSIs for Windows using CircleCI

neilalexander:
This stops using Appveyor Windows runners and instead builds Windows MSI installers using `wixl` and `msitools` on Linux in CircleCI. This means we can get rid of one of our CI pipelines.

It also tries to build the Windows binaries in the MSIs with PIE/ASLR.

Finally, it updates Wintun to the latest version and the new DLL distribution rather than MSM.


Evengard:
You definitely need to update https://yggdrasil-network.github.io/builds.html, as the link leads to a broken build now in AppVeyor.

r4sas:
@neilalexander, I hope you wouldn't delete whole project on appveyor, because in that case you will broke chocolatey package for current release.

PR #773, opened 2021-03-24
Update README, added MSI build command

vikulin:
1. Added README change


neilalexander:
Which version of Go did you test this with and why do you want to do this?

My understanding was that the PIE builds were working fine in CI using Go 1.16.

vikulin:
> 
> 
> Which version of Go did you test this with and why do you want to do this?
> 
> My understanding was that the PIE builds were working fine in CI using Go 1.16.

Done. I was building it for Go 1.14. Now it's fine. Corrected subject and description.

neilalexander:
Closing as out-of-date.

Issue #754, opened 2021-01-16
Configurable TCP congestion control algorithm


pfactum:
Currently, BBR is a hard-coded choice, but given there's [BBRv2 coming](https://github.com/google/bbr/commits/v2alpha) it would be nice to have a possibility to choose. FWIW, it is already available in the Liquorix and pf-kernel Linux forks.

What's suggested is having a knob in the config file for the user to choose the congestion control algorithm and not limiting Yggdrasil to BBR only.

ShadowJonathan:
Wouldn't different congestion algorithms start to clash with eachother and produce weird behaviour?

pfactum:
FWIW, I'm totally for ditching this option altogether in favour of using system-wide knob.

link2xt:
> FWIW, I'm totally for ditching this option altogether in favour of using system-wide knob.

+1 on this, tweaking TCP congestion control is usually a bad idea, does more harm than good when users follow the advice of "TCP congestion control comparison" articles and select some weird experimental congestion control algorithm like YeAH TCP that are just overly aggressive and win "benchmarks" because of that. Just stick to the default (CUBIC for now) and switch to BBR/BBRv2 when/if it becomes Linux default.

neilalexander:
v0.5 will no longer request BBR. We will leave the TCP congestion control algorithm up to the host.

PR #776, opened 2021-04-07
tcp_linux: make tcp congestion control algorithm selectable

pfactum:
This should address RFE in #754.

Signed-off-by: Oleksandr Natalenko <oleksandr@natalenko.name>


neilalexander (commented):
Somewhat wondering if this is a portable approach that can be used on other (non-Linux) platforms?

Might also make sense to use the `omitempty` tag and set an empty default value for other platforms that don't support it, so that we can not add unnecessary configuration options for non-supporting platforms.

pfactum:
@neilalexander please check a new push, I've employed `omitempty` as suggested.

As for other platforms, I think this may work with FreeBSD as far as I can google, but I'm not sure if FreeBSD provides BBR by default. If there's someone who uses yggdrasil on other platforms and is interested in playing with TCP congestion algorithms, this concern can be addressed in subsequent patches.

PR #781, opened 2021-05-09
Main v0.4 routing changes

Arceliar:
This is a *very early* draft of the main changes that will go into v0.4, assuming no major problems are found. The routing code has been rewritten as a separate project, which ygg imports, and then the main ygg / application code can focus on things like connection management, configuration, and the IP-specific parts (tuntap etc).

As of writing, a lot of things (API/admin socket especially) are commented out or will `panic("TODO")` if called, and the code needs some performance optimization (including the upstream package that does the routing). That said, it compiles and runs OK in the netns test, so it *technically* works.

Partial TODO list:
- [x] Clean up core code (remove obsolete things, rewrite things that no longer make much sense)
- [x] Fix core API (requires exporting some info from external routing package, so I need to do that first)
- [x] Clean up config (remove obsolete options, e.g. now only one permanent set of keys, ed25519)
- [x] ~Fix session firewall (needs to work in terms of ed keys)~
- [x] ~Test session firewall, possibly rename (it drops packets at the tuntap level, internally a crypto session may still be created for incoming traffic, if they happened to somehow know your key)~
- [x] Remove session firewall. It can't prevent memory from being allocated, so there's no reason to use this instead of an OS firewall (access controls are out of scope for this project)
- [x] Fix allowed ~encryption~ public key peering options
- [x] Fix MTU (send icmp too big message back over the network, instead of locally? so we don't need to track track MTU per remote node)
- [x] ~Either fix or~ remove CKR ~(I'd lean towards removal, other solutions exist to tunnel arbitrary IP over ygg, e.g. GRE or any VPN, and I don't like that it can lead to ygg-over-ygg problems out of the box)~ (removed, out of scope for the project in general and this PR in particular, for now you should use GRE or something else off-the-shelf instead)
- [x] NodeInfo (no longer makes sense in core... move to tuntap as a separate packet type?)

EDIT: Everything looks good from my side, ready to merge and release whenever @neilalexander is also happy with it (and has time to do all the release work)


vikulin:
Tested linux x86_64 build which based on future branch. As result found that overlay peering with tcp://<host>:port does not work. Also the build uses 0.3.16 version which obviously should be corrected.

neilalexander:
> As result found that overlay peering does not work

Overlay peering?

neilalexander:
> yes

Need you to be more specific — when you say "overlay peering", do you mean peering over Ygg (which you shouldn't be doing anyway), or just to an internet address?

vikulin:
Rebuilding...

vikulin:
You can subscribe to Telegram channel to receive latest builds:
https://t.me/yggdrasil_android

vikulin:
Tested https://github.com/yggdrasil-network/yggdrasil-go/commit/05caf36f4e4f96a9b7ff557224024cbb75598dd6 build:
1. Used the same config as in bb92e61 test
2. version is still showing 0.3.16
3. no ping is working

vikulin:
Double tested https://github.com/yggdrasil-network/yggdrasil-go/pull/781/commits/577b7118ad336f9364b73698457f52b4028514c7 build.
Ping works. Continue testing.

Vort:
Do developers plan to create test network for new version of Yggdrasil?
So anyone can connect to public peer and make tests.

vikulin:
> 
> 
> Do developers plan to create test network for new version of Yggdrasil?
> So anyone can connect to public peer and make tests.

I think Yggdrasil community has all power to create the test network without asking devs to create it.

Arceliar:
We may set up a test net after things stabilize. It's likely that there will still be backwards incompatible changes between now and the final v0.4 release, so trying to set up a test net would probably do more harm than good (since there could be silent version incompatibilities between nodes in the test net).

neilalexander (approved):

PR #791, opened 2021-06-21
Fix typo in whitepaper

bf:
"Normally, a root node will update their sequence number for frequently than this" -> "more frequently than this"


neilalexander:
Closing as the `doc` folder is no longer in the main repo. The documents in it were all stale as of v0.4.

Issue #573, opened 2019-10-10
Module tun not found in OpenVZ kernel 2.6.32-042stab127.2


Mikaela:
Yggdrasil stops being able to start on Debian 9 in OpenVZ with kernel 2.6.32… after upgrade to 0.3.10. Earlier versions didn't have issues.

```
Oct 10 20:50:24 relpda systemd[1]: [/etc/systemd/system/yggdrasil.service:11] Unknown lvalue 'CapabilityBoundSet' in section 'Service'
Oct 10 20:50:24 relpda systemd[1]: yggdrasil.service: Failed to set invocation ID on control group /system.slice/yggdrasil.service, ignoring: Oper
ation not permitted                                                                  
Oct 10 20:50:24 relpda systemd[1]: Starting yggdrasil...
Oct 10 20:50:24 relpda yggdrasil[8092]: modprobe: ERROR: ../libkmod/libkmod.c:514 lookup_builtin_file() could not open builtin file '/lib/modules/
2.6.32-042stab127.2/modules.builtin.bin'
Oct 10 20:50:24 relpda yggdrasil[8092]: modprobe: FATAL: Module tun not found in directory /lib/modules/2.6.32-042stab127.2
Oct 10 20:50:24 relpda systemd[1]: yggdrasil.service: Control process exited, code=exited status=1
Oct 10 20:50:24 relpda systemd[1]: Failed to start yggdrasil.
Oct 10 20:50:24 relpda systemd[1]: yggdrasil.service: Unit entered failed state.
Oct 10 20:50:24 relpda systemd[1]: yggdrasil.service: Failed with result 'exit-code'.
```

Mikaela:
Arceliar suggested a workaround of removing the modprobe line as a workaround and that works.

https://github.com/yggdrasil-network/yggdrasil-go/blob/045a24d74ed06f6aa658a131f3980fb8e60938b5/contrib/systemd/yggdrasil.service#L12

Arano-kai:
According to [ExecStart](https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=) syntax, prefixes may combine in any order, but it also may depends on systemd version (not sure).
Please, try to override with:
```
ExecStartPre=
ExecStartPre=+-/sbin/modprobe tun
```
and describe result.

Mikaela:
I don't understand what you are asking me to do.

Arano-kai:
```
# systemctl edit yggdrasil.service
[...]

PR #797, opened 2021-07-02
Fix yggdrasil.service file for systemd

Chaz6:
There are a couple of superfluous characters which case a warning with systemd.


Chaz6:
Closed as this is actually correct syntax as per https://github.com/yggdrasil-network/yggdrasil-go/issues/573. CentOS 7 systemd may be too old.

PR #801, opened 2021-07-06
Preallocate memory when deriving address from key

tdemin:
This makes `src/address.AddrForKey` preallocate 128 bytes for `temp` before starting the address derivation. As [benches](https://git.tdem.in/tdemin/syg_go/src/commit/f5ca266cf10d01faf60a775f8a69e003f5bb5798/crypto_test.go#L35) versus the [version that already does that](https://git.tdem.in/tdemin/syg_go/src/commit/f5ca266cf10d01faf60a775f8a69e003f5bb5798/crypto.go#L31) in syg_go show, reallocating `temp` each time in the loop takes ~ 20% of the function runtime.

The change has no impact on function behaviour.

Benchmark results

### DIFF
--- a/src/address/address.go
+++ b/src/address/address.go
@@ -64,7 +64,7 @@ func AddrForKey(publicKey ed25519.PublicKey) *Address {
 		buf[idx] = ^buf[idx]
 	}
 	var addr Address
-	var temp []byte
+	var temp = make([]byte, 0, 32)
 	done := false
 	ones := byte(0)
 	bits := byte(0)


neilalexander (approved):

src/address/address.go
```diff
@@ -64,7 +64,7 @@ func AddrForKey(publicKey ed25519.PublicKey) *Address {
 		buf[idx] = ^buf[idx]
 	}
 	var addr Address
-	var temp []byte
+	var temp = make([]byte, 0, 128)
```
Arceliar, line 67: There are better ways to implement this function if we really cared about performance. I guess it doesn't hurt anything to keep the current implementation and preallocate the slice, but I think this should be 32 bytes (since temp is at most the length of the public key, which is 32 bytes).

Arceliar (changes requested):
See: inline comment.

src/address/address.go
tdemin: True; I miscounted that for whatever reason thinking those were bits, and not bytes. :D

tdemin:
...a-a-and of course commit message said 128 bits. :-D

Arceliar (approved):

Issue #802, opened 2021-07-06
Make possible to run without root; socket creation.


zander:
The default location of the admin socket is `/var/run/yggdrasil.sock`.

There are two problems with that;

1. it is created and deleted once per session instead of once and then reused.
2. the `/var/run` dir is only writable by root.

Those two together means that on the topic of the socket, you need to run yggdrasil as root. And that is unfortunate for server software. Security wise there is a wish to avoid running things as root.

The easiest solution is to create a yggdrasil subdir in /var/run which avoids the permissions issue. Not sure how that works together with #449.  
The alternative solution is to not delete the socket file upon exit, as that avoids the requirement to be root to do so.  Edit; this means the app can drop root privs the moment the socket is created, or use a separate method to create it.

neilalexander:
This is being looked at in #927.

PR #804, opened 2021-07-06
Add -uid and -uid flags to drop permissions

tsuraan:
yggdrasil seems to prefer to run as a privileged user, but it also seems
to work fine if permissions are dropped after the socket initialization
is performed. This adds -uid and -gid flags so that an instance run with
root perms can drop them once it's ready.

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -193,6 +193,8 @@ type yggArgs struct {
 	getaddr       bool
 	getsnet       bool
 	loglevel      string
+	runuid        int
+	rungid        int
 }
 
 func getArgs() yggArgs {
@@ -207,7 +209,10 @@ func getArgs() yggArgs {
 	getaddr := flag.Bool("address", false, "returns the IPv6 address as derived from the supplied configuration")
 	getsnet := flag.Bool("subnet", false, "returns the IPv6 subnet as derived from the supplied configuration")
 	loglevel := flag.String("loglevel", "info", "loglevel to enable")
+	runuid := flag.Int("uid", -1, "drop privileges to this user id")
+	rungid := flag.Int("gid", -1, "drop privileges to this group id")
 	flag.Parse()
+
 	return yggArgs{
 		genconf:       *genconf,
 		useconf:       *useconf,
@@ -220,6 +225,8 @@ func getArgs() yggArgs {
 		getaddr:       *getaddr,
 		getsnet:       *getsnet,
 		loglevel:      *loglevel,
+		runuid:        *runuid,
+		rungid:        *rungid,
 	}
 }
 
@@ -364,6 +371,17 @@ func run(args yggArgs, ctx context.Context, done chan struct{}) {
 	address := n.core.Address()
 	subnet := n.core.Subnet()
 	public := n.core.GetSelf().Key
+	// Lower permissions from root to something else, if the user wants to
+	if syscall.Getuid() == 0 {
+		if args.rungid > 0 {
+			fmt.Println("Dropping gid to ", args.rungid)
+			syscall.Setgid(args.rungid)
+		}
+		if args.runuid > 0 {
+			fmt.Println("Dropping uid to ", args.rungid)
+			syscall.Setuid(args.runuid)
+		}
+	}
 	logger.Infof("Your public key is %s", hex.EncodeToString(public[:]))
 	logger.Infof("Your IPv6 address is %s", address.String())
 	logger.Infof("Your IPv6 subnet is %s", subnet.String())


tsuraan:
So, I have no idea whether this is desired, or if there's a better way to do it, or what, but it doesn't look like ygg needs to have so many permissions once it's up and running. Might be worth looking into, or maybe not.

tdemin:
@tsuraan you know this makes Yggdrasil not portable to other OSes, right?

cmd/yggdrasil/main.go
```diff
@@ -364,6 +371,17 @@ func run(args yggArgs, ctx context.Context, done chan struct{}) {
 	address := n.core.Address()
 	subnet := n.core.Subnet()
 	public := n.core.GetSelf().Key
+	// Lower permissions from root to something else, if the user wants to
+	if syscall.Getuid() == 0 {
+		if args.rungid > 0 {
+			fmt.Println("Dropping gid to ", args.rungid)
+			syscall.Setgid(args.rungid)
```
tdemin, line 378: You know this makes Yggdrasil no longer portable, right?

tdemin (changes requested):

cmd/yggdrasil/main.go
tsuraan: Yeah, I see the circleci failure under the "other" OS (I assume that's windows?). I'm a bit surprised there's an OS that doesn't have UIDs, but just a little bit. Anyhow, I guess the fix for that would be to move the args struct and arg parser into a arg_windows.go and a arg_posix.go, and only have the gid/uid options in the args_posix side. Similarly define a drop_perm_unix.go that does the above, and a drop_perm_windows.go that's just a no-op. I can definitely do that, but before I do, is there interest in privilege dropping in yggdrasil? I'm happy to do the grunt work, but if it's something that's not seen as necessary anyhow, then I can just drop this.

Arceliar: More importantly, `syscall.Setgid` (or setting UIDs) probably won't work in Go. Or to be exact, it will set the ID for whatever random thread the goroutine happened to be running on at the time.

The language recently added https://golang.org/pkg/syscall/#AllThreadsSyscall which is meant to address issues like this, but it has some caveats associated with it.

Arceliar: If I understand correctly, the `AllThreadsSyscall` function is only available on linux, so this feature wouldn't be cross-platform.

tsuraan: POSIX setuid is required to set the uid for the entire process ([ref](https://pubs.opengroup.org/onlinepubs/000095399/functions/setuid.html) and a [stackoverflow](https://stackoverflow.com/questions/45397316/is-this-possible-to-switch-the-user-and-group-of-an-application-thread)). It definitely appears to be working under FreeBSD and Linux, but I haven't tried it under OSX. I'd expect it to work the same there, as OSX's POSIX layer appears to still be [based on FreeBSD](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/KernelProgramming/BSD/BSD.html).

Or am I misunderstanding the problem? It looks like Linux is actually the weird one, because it does (in the kernel) implement per-thread uids, and sane programs depend on glibc or musl to provide the POSIX semantics that people tend to expect ([more](https://ewontfix.com/17/)). If Go's "libc" really is just sending the raw syscalls to the kernel, then I can see how the AllThreadsSyscall would be needed to be run explicitly on LInux systems. Unfortunately I'm not actually a Go programmer, and I have no idea what AllThreadsSyscall's function signature means. So, I guess I'll drop this, unless somebody really wants to see it completed.

zander:
Related https://github.com/yggdrasil-network/yggdrasil-go/issues/802

cmd/yggdrasil/main.go
Arceliar: Looking into it more, it appears that `Setgid` and `Setuid` were rewritten to use `AllThreadsSyscall` on linux (as long as `cgo` isn't being used), so this should be doing the right thing.

Some refactoring/cleanup is needed to avoid breaking things on platforms where `syscall.Setuid`/`syscall.Setgid` are undefined (if any... windows maybe?). That shouldn't be too difficult, I can probably take a look at in the next few days.

tsuraan: If somebody else could figure out how to do the conditional compilation thing, that would be cool. I tried to split out main.go's argv into its own main_posix.go and main_other.go with `// +build` directives for per-OS compilation, and I just couldn't get the thing to compile. I don't know Go though, and its package system seems to be pretty unique, so I'm sure I was just doing something really basic wrong.

neilalexander:
Closing as the approach in #927 is a bit more portable.

PR #805, opened 2021-07-06
Update macOS installer and launchd file

russelldavies:
This commit makes a few improvements and changes file paths to more conventional ones.


russelldavies:
Looks like the build failed because it took too long to install Go. Or is this is symptom of the changes breaking something?

neilalexander:
In principle changing the paths is OK, but since existing installations will have their config at `/etc/yggdrasil.conf`, we need to make sure we’re moving that rather than abandoning it for a new config in `/usr/local/etc`.

russelldavies:
Good point. I've amended the commit to handle the existing file path.

PR #812, opened 2021-07-20
Make message clearer and downgrade

zander:
This error message was shown for incoming connections which is made clearer with the changed message
and after a chat with neilalexander he suggested to downgrade it to debug as it doesn't make sense to litter the log with them.

### DIFF
--- a/src/core/link.go
+++ b/src/core/link.go
@@ -169,7 +169,7 @@ func (intf *link) handler() (chan struct{}, error) {
 		return nil, errors.New("failed to decode metadata")
 	}
 	if !meta.check() {
-		intf.links.core.log.Errorf("Failed to connect to node: %s is incompatible version (local %s, remote %s)",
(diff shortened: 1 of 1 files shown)


src/core/link.go
```diff
@@ -169,7 +169,7 @@ func (intf *link) handler() (chan struct{}, error) {
 		return nil, errors.New("failed to decode metadata")
 	}
 	if !meta.check() {
-		intf.links.core.log.Errorf("Failed to connect to node: %s is incompatible version (local %s, remote %s)",
+		intf.links.core.log.Debugf("Rejected incoming connection: %s is incompatible version (local %s, remote %s)",
```
Arceliar, line 172: This shouldn't read "incoming", since it will display the same message for any connection (outgoing or incoming). I think "Rejected connection" would be good enough.

Arceliar (changes requested):

neilalexander (approved):
Looks reasonable if @Arceliar is happy with this.

PR #813, opened 2021-07-21
Allow yggdrasil bind to ports <1024

### LINKED ISSUE #809: systemd service file broken
Hello there!

I just installed yggdrasil 0.4.0 on a debian server. Running `yggdrasil -useconffile /etc/yggdrasil.conf` as root worked just fine but the systemd service just wouldn't start:
```
panic: listen tcp 0.0.0.0:443: bind: permission denied
```

I'm running systemd version `241-7~deb10u7` on `Debian GNU/Linux 10 (buster)` with kernel `Linux 4.19.0-16-amd64 #1 SMP Debian 4.19.181-1 (2021-03-19)`.

After looking around on the internet I came across a [similar problem in a different software](https://serverfault.com/a/899964) and that gave me a clue how to fix it. In `/etc/systemd/system/yggdrasil.service` (which is [this file in this repo](https://github.com/yggdrasil-network/yggdrasil-go/blob/master/contrib/systemd/yggdrasil.service)) I had to append `CAP_NET_BIND_SERVICE` to the line starting with `CapabilityBoundingSet`.

Line before editing:
```
CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_RAW
```

Line after editing:
```
CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_RAW CAP_NET_BIND_SERVICE
```

That fixed the issue for me and the systemd service was able to start.

I don't know if that is the proper way of fixing this issue or if something different would be a more elegant approach. So I won't be filing a PR for this. I just wanted to let you know about the problem and how it can be fixed.

cofob:
Fixes #809


Arceliar (commented):
This looks like it's probably fine, but I'm wary of anything that touches the systemd files... so I'll defer to @neilalexander on this one.

zander:
The default port is > 1024, which does not require the permission and it would not be a good idea to require the permission for the majority that does not need it. The term "broken" as used in #809 does seem to be suggesting that too secure is a bad feature. I personally think an Internet service that is growing in popularity will benefit from the exact opposite approach. If you can edit your ygg config, you can edit your service file.

What might be nice is to add a comment indicating a solution for those that want it, and definitely beneficial would be to have a note in the installation documentation. Both would IMO solve the problem without decreasing security.

cofob:
> The default port is > 1024, which does not require the permission and it would not be a good idea to require the permission for the majority that does not need it. The term "broken" as used in #809 does seem to be suggesting that too secure is a bad feature. I personally think an Internet service that is growing in popularity will benefit from the exact opposite approach. If you can edit your ygg config, you can edit your service file.
> 
> What might be nice is to add a comment indicating a solution for those that want it, and definitely beneficial would be to have a note in the installation documentation. Both would IMO solve the problem without decreasing security.

Yggdrasil does not have a "standard" port for configuring the "Listen" directive.  
This will solve future Issues, and I don't think allowing bindings to ports lower than 1024 will lower the security of the system.

Adding a comment with a possible solution of the problem is also a good idea, but why leave the problem when it is so easy to solve?

neilalexander (commented):
I’m in two minds about this. On the one hand I don’t think there’s really a great risk to doing this, but it’s still  “fixing” a problem that just doesn’t exist in the default configuration. Are there compelling reasons for binding to ports lower than 1024?

I think there are plenty of people who would argue that running with the least number of capabilities is generally a better security model to subscribe to by default, and anyone who wants to change it for their own configuration should be free to do so. I seem to recall systemd has a way of “extending” service files so that you can persist changes across package upgrades but I’d need to look into how that works.

My other, albeit slightly lesser concern is that it creates a double-standard across the packages out there — this will “fix” the behaviour for distributions using our systemd service unit but does nothing anywhere else, e.g. OpenRC or others.

neilalexander:
This is what I was thinking of:

> Along with a unit file foo.service, a "drop-in" directory foo.service.d/ may exist. All files with the suffix ".conf" from this directory will be merged in the alphanumeric order and parsed after the main unit file itself has been parsed. This is useful to alter or add configuration settings for a unit, without having to modify unit files. Each drop-in file must contain appropriate section headers. For instantiated units, this logic will first look for the instance ".d/" subdirectory (e.g. "foo@bar.service.d/") and read its ".conf" files, followed by the template ".d/" subdirectory (e.g. "foo@.service.d/") and the ".conf" files there. Moreover for unit names containing dashes ("-"), the set of directories generated by repeatedly truncating the unit name after all dashes is searched too. Specifically, for a unit name foo-bar-baz.service not only the regular drop-in directory foo-bar-baz.service.d/ is searched but also both foo-bar-.service.d/ and foo-.service.d/. This is useful for defining common drop-ins for a set of related units, whose names begin with a common prefix. This scheme is particularly useful for mount, automount and slice units, whose systematic naming structure is built around dashes as component separators. Note that equally named drop-in files further down the prefix hierarchy override those further up, i.e. foo-bar-.service.d/10-override.conf overrides foo-.service.d/10-override.conf.

zhoreeq:
> I’m in two minds about this. On the one hand I don’t think there’s really a great risk to doing this, but it’s still “fixing” a problem that just doesn’t exist in the default configuration. Are there compelling reasons for binding to ports lower than 1024?

Yes, it can work over a restrictive corporate firewall if bind to port 443

Vort:
> Yes, it can work over a restrictive corporate firewall if bound to port 443

Makes life harder for any censor who fears to touch 443 actually.

zander:
Would https://yggdrasil-network.github.io/installation.html be the best place to document this?

rany2:
> My other, albeit slightly lesser concern is that it creates a double-standard across the packages out there — this will “fix” the behaviour for distributions using our systemd service unit but does nothing anywhere else, e.g. OpenRC or others.

OpenRC service has the ability to bind to ports <1024 because it runs as root with no restrictions. Only the systemd service has those mitigations.

neilalexander (approved):

PR #814, opened 2021-07-22
Move yggdrasilctl responses to separate functions (small refactoring)

kotovalexarian:
It's always recommended to split code into small units such as functions because this makes it more understandable. Many code linting tools ([RuboCop](https://rubocop.org) for Ruby programming language for example) have metrics which warn about too long functions: Assigment-Branch-Condition size, methods length, etc. Also this opens the opportunity of testing the code in future.

PR #815, opened 2021-07-22
Split yggdrasilctl code into separate functions (refactoring)

kotovalexarian:
It's always recommended to split code into small units such as functions because this makes it more understandable. Many code linting tools ([RuboCop](https://rubocop.org) for Ruby programming language for example) have metrics which warn about too long functions: Assigment-Branch-Condition size, methods length, etc. Also this opens the opportunity of testing the code in future.

The previous PR (#814) have been accidentally closed by GitHub when I renamed the branch.

### DIFF
--- a/cmd/yggdrasilctl/cmd_line_env.go
+++ b/cmd/yggdrasilctl/cmd_line_env.go
@@ -15,8 +15,8 @@ import (
 )
 
 type CmdLineEnv struct {
-	args []string
-	endpoint, server string
+	args                 []string
+	endpoint, server     string
 	injson, verbose, ver bool
 }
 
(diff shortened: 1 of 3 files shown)


neilalexander (changes requested):
I absolutely welcome the split to multiple functions but I feel like the `cmd_line_env` package is an unnecessary point of complexity. I'd prefer it if that code could be put back into the `main` package.

kotovalexarian:
> I absolutely welcome the split to multiple functions but I feel like the `cmd_line_env` package is an unnecessary point of complexity. I'd prefer it if that code could be put back into the `main` package.

Ok, I'll change the code to follow this.

kotovalexarian:
@neilalexander done

neilalexander (approved):

neilalexander:
Thanks for this!

PR #816, opened 2021-07-24
[Security] do not run as root.

zander:
An Internet accessible service should aim to have as little as
possible attack surface, which is much easier to do when running
with the absolute minimum number of priviledges at all times.

This makes the systemd setup run the service as a user 'yggdrasil'
and uses the systemd feature `RuntimeDirectory` to auto-create
the /var/run/yggdrasil dir so our non-elevated client can still
create the socket. (see #802)

The sysusers file will cause the user be created on first install
using the sysusers subsystem, this can be used by packagers as well.

### DIFF
--- a/contrib/systemd/yggdrasil.sysusers
+++ b/contrib/systemd/yggdrasil.sysusers
@@ -1,0 +1,1 @@
+u yggdrasil - "Yggdrasil network daemon"
--- a/contrib/systemd/yggdrasil.sysusers
+++ b/contrib/systemd/yggdrasil.sysusers
@@ -1,1 +1,0 @@
-u yggdrasil - "Yggdrasil network daemon"
--- a/src/defaults/defaults_linux.go
+++ b/src/defaults/defaults_linux.go
@@ -8,7 +8,7 @@ package defaults
 func GetDefaults() platformDefaultParameters {
 	return platformDefaultParameters{
 		// Admin
-		DefaultAdminListen: "unix:///var/run/yggdrasil.sock",
+		DefaultAdminListen: "unix:///var/run/yggdrasil/yggdrasil.sock",
 
 		// Configuration (used for yggdrasilctl)
 		DefaultConfigFile: "/etc/yggdrasil.conf",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - in case of vulnerabilities.
 -->
 
+## [future]
+
+### Changed
+- The `AdminListen` option and `yggdrasilctl` now default to `unix:///var/run/yggdrasil/yggdrasil.sock` on Linux
+
+
 ## [0.4.2] - 2021-11-03
 ### Fixed
 - Reverted a dependency update which resulted in problems building with Go 1.16 and running on Windows
--- a/contrib/systemd/yggdrasil.service
+++ b/contrib/systemd/yggdrasil.service
@@ -6,8 +6,6 @@ After=network-online.target
 After=yggdrasil-default-config.service
 
 [Service]
-User=yggdrasil
-Group=yggdrasil
 ProtectHome=true
 ProtectSystem=true
 SyslogIdentifier=yggdrasil
@@ -16,6 +14,12 @@ ExecStart=/usr/bin/yggdrasil -useconffile /etc/yggdrasil.conf
 ExecReload=/bin/kill -HUP $MAINPID
 Restart=always
 TimeoutStopSec=5
+Group=yggdrasil
+User=yggdrasil-dyn
+DynamicUser=true
+ProtectSystem=strict
+NoNewPrivileges=true
+ReadWritePaths=/var/run/yggdrasil /run/yggdrasil
 
 # make sure /var/run/yggdrasil/ is created writable for the user.
 RuntimeDirectory=yggdrasil
--- a/contrib/systemd/yggdrasil.service
+++ b/contrib/systemd/yggdrasil.service
@@ -6,16 +6,21 @@ After=network-online.target
 After=yggdrasil-default-config.service
 
 [Service]
+User=yggdrasil
 Group=yggdrasil
 ProtectHome=true
 ProtectSystem=true
 SyslogIdentifier=yggdrasil
-CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_RAW CAP_NET_BIND_SERVICE
 ExecStartPre=+-/sbin/modprobe tun
 ExecStart=/usr/bin/yggdrasil -useconffile /etc/yggdrasil.conf
 ExecReload=/bin/kill -HUP $MAINPID
 Restart=always
 TimeoutStopSec=5
+
+# make sure /var/run/yggdrasil/ is created writable for the user.
+RuntimeDirectory=yggdrasil
+# the small list of admin capabilities we need to do our job
+AmbientCapabilities=CAP_NET_ADMIN CAP_NET_RAW CAP_NET_BIND_SERVICE
 
 [Install]
 WantedBy=multi-user.target


cofob (approved):
Looks fine

src/defaults/defaults_linux.go
```diff
@@ -7,7 +7,7 @@ package defaults
 func GetDefaults() platformDefaultParameters {
 	return platformDefaultParameters{
 		// Admin
-		DefaultAdminListen: "unix:///var/run/yggdrasil.sock",
+		DefaultAdminListen: "unix:///var/run/yggdrasil/yggdrasil.sock",
```
Arceliar, line 10: This breaks running if the `/var/run/yggdrasil` folder doesn't exist, so the default depends on the systemd package (or otherwise creating the folder). The binary produced by `./build` needs to be able to run with default settings (i.e. `-autoconf`) with no further changes to the OS/environment.

Arceliar (changes requested):

src/defaults/defaults_linux.go
zander: I didn't test this and indeed if yggdrasil does not try to create the directory in which it wants to place the socket (which IMOHO is a bug) then this will fail.

An actual go progammer is required to make ygg create the directory before it attempts to create a file in it (assuming this doesn't work as you state). I am not a Go programmer, would appreciate a patch. I guess its not too complex.

OlegGirko: No problem, default socket path can be different if run with `-autoconf`. Probably, it should be created in user's home directory if run this way.

zander: good point. Likely best to use /run/user/{UID}/ which is where Linux by default makes those.

For instance wayland creates its sockets there.

OlegGirko: No, it's bad to hardcode this path. The better way would be to use contents `XDG_RUNTIME_DIR` environment variable if it's present, otherwise something like `.yggdrasil.sock` in user's home directory. I presume, this will be a good default for those who still insist on not using `systemd`.

OlegGirko: Actually, defaults should be dependent on whether config is generated by regular user or `root` (not on `-autoconf` option).

Regular user should have everything in the home directory (and socket `$XDG_RUNTIME_DIR` if this variable is present).
But for `root` user everything should be in LSB-conforming directories (config in `/etc/yggdrasil`, socket in `/run/yggdrasil`).

contrib/systemd/yggdrasil.sysusers
```diff
@@ -0,0 +1 @@
+u yggdrasil - "Yggdrasil network daemon"
```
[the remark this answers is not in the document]
zander: By having it in the contrib dir it can be used by many distro packagers.

src/defaults/defaults_linux.go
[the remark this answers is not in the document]
zander: comment is duplicate of what Arceliar wrote.

contrib/systemd/yggdrasil.sysusers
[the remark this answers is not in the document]
zhoreeq: You broke the Debian package by not including this file in contrib/deb/generate.sh

zander:
To followup after the discussion on matrix.

* the executable as compiled by `build` should, when using -autoconf use the XDG_RUNTIME_DIR (typically /run/user/UID) as a basdir instead. The rationale here is that services (daemons in Linux) do not have this variable set and thus finding this env-var means its started by a user.  The `yggdrasilctl` executable can follow the same logic and thus find the socket.
* the package-managers will be very happy to reuse the files this PR updates, they lack the XDG env-var and thus will make the file go to `/run/yggdrasil/`. Packagers for distro's not using systemd will need a packager-note to make clear that need to create this dir.
* maybe users are used to running this as root, as such I suggest that yggdrasil auto-creates the directory the user requested the socket to be inserted in. This code will only hit when the two above cases are not used as they ensure the dir to exist.

I would be very sad if yggdrasil would turn off the admin interface before there is a replacement as that makes deployment and sysadmin-jobs much harder.

Arceliar:
One more thing to add:

Running as an unprivileged user means the old `AdminListen` value (`/var/run/yggdrasil.sock`) will need to be recognized and replaced when updating. Otherwise, updating the package (via `apt-get upgrade` or similar) will leave the system in a broken state. The new systemd file would try to run ygg as an unprivileged user, but using a config file with the old (root-only) path, which leads to a runtime crash. This is true regardless of what the new default value is.

zander:
> Running as an unprivileged user means the old AdminListen value (/var/run/yggdrasil.sock) will need to be recognized and replaced when updating.

Yes, this is one of those problems: the ideal time to do this? Before 0.4.0,  the second best time to do this? Now.

The solution there is for packagers to add an 'upgrade' line. Ygg does not have to do anything as the problem does not exist for people using its compiled binary.
`perl -spi.orig -e 's/(AdminListen: unix:\/\/\/var\/run\/)(yggdrasil.sock)/$1\yggdrasil\/$2/;' /etc/yggdrasil.conf`

> We can make the AdminListen something that's configurable at compile time. The default value would be "none" for all platforms, but the package maintainers could override. 

That is possible. Sounds complicated compared to simply doing a makedir.  
You might consider instead having a new argument that is used in combination wih `--genconf` to create a useful line for AdminListen.

This has the advantage that your second systemd config line is the one that actually calls genconf and thus this change has lower impact.

kotovalexarian:
Why not just drop root privileges after socket file setup?

jgoerzen:
Here are some suggestions to make this work better:

Don't create a yggdrasil sysusers file, and instead use the DynamicUser option:

```
User=yggdrasil-dyn
Group=yggdrasil
DynamicUser=true
```

(Note: User must not match yggdrasil here, since it isn't defined statically)

Also, set:

```
ProtectSystem=strict
NoNewPrivileges=true
ReadWritePaths=/var/run/yggdrasil /run/yggdrasil
```

ProtectSystem=strict prevents it from modifying basically anything on the filesystem, so therefore we have to specify the possible locations for the runtime directory under ReadWritePaths.

zander:
@jgoerzen sounds like an interesting idea.  Seems to make the systemd deployment one file, which is nice.

Would you be able to comment on the minimum version of systemd needed for all the features your idea uses?

jgoerzen:
Sure.  Per the systemd NEWS file:

- DynamicUser was added in version 232, 2016-11-03
- NoNewPrivileges was introduced before NEWS started; it was already mentioned as existing in v227, 2015-10-07
- ReadWritePaths was also introduced before NEWS started; it was already mentioned as existing in v231, 2016-07-25
- ProtectSystem was already in use, and was introduced in v214, 2014-06-11

As far as I can tell, this is supported by all major distributions that are under active support.

Versions of systemd in Debian that are still under standard or Long-Term Support (LTS):
- stretch (June 2017): 232
- buster (July 2019): 241
- bullseye (August 2021): 247
- Source: https://packages.debian.org/search?keywords=systemd

Versions of systemd in Ubuntu that are still under general support:
- bionic (April 2018): 237
- focal (April 2020): 245
- hirsute (April 2021): 247
- impish (October 2021: 248
- Source: https://packages.ubuntu.com/search?keywords=systemd

Versions of systemd in Fedora that are still under general support:
- Fedora 34: 248
- Fedora 35: 249
- Source: https://packages.fedoraproject.org/pkgs/systemd/systemd/

RHEL/CentOS forks RockyLinux and AlmaLinux are both shipping version 239, as did CentOS 8.

Various version comparisons can be found at https://pkgs.org/download/systemd

zander:
I tested this new idea on my archlinux (i.e. latest) and also on my server which is debian old-stable (v 9).
It works great on both;

`systemctl --version` gives;

`systemd 249 (249.7-2-arch)` and `systemd 232` respectively

I'll push an update shortly.

jgoerzen:
An update:

Per discussion in Matrix, CAP_NET_RAW is no longer necessary.

You may be interested in the service file I'm shipping with the package that will go into Debian; it's linked at https://github.com/yggdrasil-network/yggdrasil-go/issues/874#issuecomment-1017643568

neilalexander:
I removed `CAP_NET_RAW` already in yggdrasil-network/yggdrasil-go@a4bdf3de3266d5a248a11e060796e817a33c3213.

neilalexander:
Closing in favour of #927.

PR #817, opened 2021-07-24
[revised] Add -uid and -uid flags to drop permissions

Arceliar:
This is built on #804 (thanks @tsuraan)

It's essentially identical to that PR, except it wraps the syscalls in a separate function, so we can return an error on platforms where this isn't supported (which seems to at least include windows).

I'm not sure exactly what build directives make the most sense here... at the moment, it only supports `linux`, `darwin`, `freebsd`, and `openbsd`, but maybe `!windows` is more correct.

We should probably think a bit about what behavior we want. If we always want to drop permissions, then we could maybe skip the `-uid` and `-gid` args, and hard code a specific user to switch to (`nobody` or whatever makes the most sense).

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -376,12 +376,18 @@ func run(args yggArgs, ctx context.Context, done chan struct{}) {
 	// Lower permissions from root to something else, if the user wants to
 	if syscall.Getuid() == 0 {
 		if args.rungid > 0 {
-			fmt.Println("Dropping gid to ", args.rungid)
-			syscall.Setgid(args.rungid)
+			logger.Infoln("Setting gid to:", args.rungid)
+			if err := setgid(args.rungid); err != nil {
+				logger.Errorln("Failed to set gid:", err)
+				return
+			}
 		}
 		if args.runuid > 0 {
-			fmt.Println("Dropping uid to ", args.rungid)
-			syscall.Setuid(args.runuid)
+			logger.Infoln("Setting uid to:", args.runuid)
+			if err := setuid(args.runuid); err != nil {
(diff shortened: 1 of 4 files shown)


zander:
The creation of the socket file is a filesystem permissions issue.

It is at best very unconventional to solve this filesystem permissions issue the way this PR suggests: starting out as superuser and trusting ygg to drop to the permissions it wants.
Honestly, this is not a clean solution to solve a simple "This user can not write in this dir" problem.

Arceliar:
This isn't just about the socket file. If the process was started as root, then there's no reason to keep running as root after the point where elevated permissions may have been needed. A [long standing issue in go](https://github.com/golang/go/issues/1435) prevented this until recently.

neilalexander:
Adding my own review to remind me to test this on macOS and FreeBSD.

cmd/yggdrasil/main.go
```diff
@@ -366,6 +373,23 @@ func run(args yggArgs, ctx context.Context, done chan struct{}) {
 	address := n.core.Address()
 	subnet := n.core.Subnet()
 	public := n.core.GetSelf().Key
+	// Lower permissions from root to something else, if the user wants to
+	if syscall.Getuid() == 0 {
+		if args.rungid > 0 {
+			logger.Infoln("Setting gid to:", args.rungid)
+			if err := setgid(args.rungid); err != nil {
+				logger.Errorln("Failed to set gid:", err)
+				return
```
zander, line 382: instead of simply returning, you likely want to abort() here.

Not being able to change gid/uid (or egid/euid) is not a failure you should ignore and then simply continue running.

neilalexander:
Closing in favour of #927.

PR #821, opened 2021-07-28
Change systemd service unit file path to /usr/lib

### LINKED ISSUE #820: Default .service unit files should reside in /usr/lib/systemd/system (.deb)
According to systemd.unit(5) manpage (https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Unit%20File%20Load%20Path), system units installed by the distribution package manager should reside in /usr/lib/systemd/system. Currently with .deb packages from http://neilalexander.s3.dualstack.eu-west-2.amazonaws.com/deb/ repository, the service unit files are installed in /etc/systemd/system. I believe it's a good idea to move the default installation path of the systemd unit files, as /etc/systemd/system is intended for administrator-placed unit files and overrides /usr/lib/systemd/system.

tkiapril:
As noted in #820, systemd service unit files installed by package managers should be residing in /usr/lib/systemd/system. This little PR makes .deb packages built with the contrib script to be in line with the manpage description. 

Resolves #820.

### DIFF
--- a/contrib/deb/generate.sh
+++ b/contrib/deb/generate.sh
@@ -38,7 +38,7 @@ echo "Building $PKGFILE"
 mkdir -p /tmp/$PKGNAME/
 mkdir -p /tmp/$PKGNAME/debian/
 mkdir -p /tmp/$PKGNAME/usr/bin/
-mkdir -p /tmp/$PKGNAME/etc/systemd/system/
+mkdir -p /tmp/$PKGNAME/usr/lib/systemd/system/
 
 cat > /tmp/$PKGNAME/debian/changelog << EOF
 Please see https://github.com/yggdrasil-network/yggdrasil-go/
@@ -68,7 +68,7 @@ EOF
 cat > /tmp/$PKGNAME/debian/install << EOF
 usr/bin/yggdrasil usr/bin
 usr/bin/yggdrasilctl usr/bin
-etc/systemd/system/*.service etc/systemd/system
+usr/lib/systemd/system/*.service usr/lib/systemd/system
 EOF
 cat > /tmp/$PKGNAME/debian/postinst << EOF
 #!/bin/sh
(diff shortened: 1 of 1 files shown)


cofob (approved):

neilalexander (commented):
While the sentiment is definitely right, I'm a little bit puzzled that on a Debian Stretch system, `/usr/lib/systemd/system` doesn't exist (although `/lib/systemd/system` does)?

tkiapril:
I researched around a bit about the separated `/usr/lib` and `/lib` issue, and it turns out that the problem stems from something called the `/usr merge`; apparently, Debian systems have had separate `/{bin,sbin,lib}` from `/usr/{bin,sbin,lib}` due to historical reasons including booting directly into minimal system on disk without having to have an initramfs. (https://wiki.debian.org/UsrMerge) However, it seems like since buster release `debootstrap` has moved to a default of merged /usr, and it has been ruled by the Debian TC in Feb 2021 that Debian `bookworm` will drop support for unmerged /usr layout (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=978636). I can confirm this on a new installation of Ubuntu 20.04.2 (will also check on Debian installations.) @neilalexander If your installation of Debian stretch has been bootstrapped before buster release, it might as well as have the former directory layout, as it seems like the merge is not automatically done. As the future releases of Debian will have the merged layout, and since the TC's advice on packaging is that you can use either layout (https://lists.debian.org/debian-devel-announce/2019/03/msg00001.html), I think it would be better to have the .service files located at /usr/lib to have it futureproof.

cobratbq:
I had noticed the same unexpected location of the *.service files. I would like to see this change, although I don't know anything about the circumstances described above.

The PR seems incomplete though: changes required on line __71__ and __117__, __118__.

Note that, in addition to the fact that the files are "expected in a different location", having the current location blocks several operations:
- overriding at-run-time configuration changes via `/run/systemd/system`
- overriding sysadmin changes via `/etc/systemd/system`
- management operations such as `systemctl mask yggdrasil`, which rely on `etc/systemd/system`

tkiapril:
Those lines somehow slipped my attention, thanks for noticing @cobratbq! Changes made to the lines.

cobratbq:
I think it is possible to query for the required path using `systemd-path systemd-system-unit`. Although on Debian Bullseye it will still point to `/lib/systemd/system` even though `/lib` is a symlink now.

tkiapril:
That's a great suggestion, albeit I'm not sure about any regression it might cause in regards to the debian packaging format since I'm not that used to debian as a whole (my primary distro is Arch.) Would it be fine to substitute all occurences of usr/lib/systemd/system on the script to $(systemd-path systemd-system-unit)?

cobratbq:
Unfortunately, neither am I.

@tkiapril Could you check if `systemd-path systemd-system-unit` points to the correct location for Arch? I would expect `/usr/lib/systemd/system`. It would help to confirm if this is indeed useful.

tkiapril:
For Arch, it indeed resolves to /usr/lib/systemd/system. However, when I tried it in CentOS 8, it says `Path systemd-system-unit not known.`. Would this be significant in some way?

neilalexander:
I'm not convinced this isn't going to break some systems thanks to the `/usr` thing so I think we'll leave the existing package as-is for now and distribution packagers (who will probably produce better quality packages than we can) can adapt if they need to.

PR #822, opened 2021-07-28
TLS Server Name Indication

neilalexander:
This PR does two things:

* Sets the TLS SNI on TLS peerings to the hostname in the peering URI (but *not* naked IP address literals)
* Allows overriding the SNI with a custom value by specifying the query string `?sni=...`

### DIFF
--- a/src/core/link.go
+++ b/src/core/link.go
@@ -103,8 +103,8 @@ func (l *links) call(u *url.URL, sintf string) error {
 		// the host-port combo from the query option and then seeing if it parses to an
 		// IP address successfully or not.
 		if sni := u.Query().Get("sni"); sni != "" {
-			if host, _, err := net.SplitHostPort(sni); err == nil && net.ParseIP(host) == nil {
-				tcpOpts.tlsSNI = host
+			if net.ParseIP(sni) == nil {
+				tcpOpts.tlsSNI = sni
 			}
 		}
 		// If the SNI is not configured still because the above failed then we'll try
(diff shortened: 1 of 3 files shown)


src/core/tls.go
```diff
@@ -77,8 +77,8 @@ func (t *tcptls) init(tcp *tcp) {
 	}
 }
 
-func (t *tcptls) configForOptions(options *tcpOptions) *tls.Config {
-	config := *t.config
+func (t *tcptls) configForOptions(options *tcpOptions, serverName string) *tls.Config {
```
Arceliar, line 80: If the `tcpOptions` contain `tlsSNI`, then I don't see a reason to pass `serverName` here... we could just make sure the field in the `tcpOptions` is correct, and read it from there (it looks like that's effectively what already happens, it just passes through the arguments as a needless extra step).

src/core/tls.go
```diff
@@ -103,11 +103,14 @@ func (t *tcptls) configForOptions(options *tcpOptions) *tls.Config {
 		}
 		return nil
 	}
-	return &config
+	if serverName != "" {
+		config.ServerName = serverName
```
Arceliar, line 107: See above, it seems like `config.ServerName = tcpOptions.tlsSNI` would unconditionally do the right thing?

src/core/link.go
```diff
@@ -98,6 +98,14 @@ func (l *links) call(u *url.URL, sintf string) error {
 		l.tcp.call(pathtokens[0], tcpOpts, sintf)
 	case "tls":
 		tcpOpts.upgrade = l.tcp.tls.forDialer
+		tcpOpts.tlsSNI = u.Query().Get("sni")
```
Arceliar, line 101: Maybe we should parse the contents of `?sni=` before assigning it to `tlsSNI`, just to make sure it's not nonsense? (E.g. not an IP literal)

Arceliar (changes requested):
Possible sanity check and minor cleanup

Arceliar (approved):

PR #828, opened 2021-08-05
Add some tests

kotovalexarian:


src/config/config_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"bytes"
+	"encoding/hex"
+	"testing"
+)
+
+func TestConfig_Keys(t *testing.T) {
+	var nodeConfig NodeConfig
+
+	if nodeConfig.PublicKey != "" {
```
neilalexander, line 12: I don’t think these assertions can ever fail when doing `var nodeConfig NodeConfig`, since the values should be their empty defaults.

src/config/config_test.go
```diff
@@ -0,0 +1,63 @@
...
+		t.Fatal("public key is not empty by default")
+	}
+
+	if nodeConfig.PrivateKey != "" {
```
neilalexander, line 16: Same here.

neilalexander (commented):
Looks mostly sane, couple minor things.

kotovalexarian:
@neilalexander Fixed.

PR #826, opened 2021-08-01
Fix fail start yggdrasil service on boot

NecroRomnt:
On system boot I was take yggdrsil errors. Service can't listen local IP because connection not upped yet.
I suggest replace network.target to network-online.target in file yggdrasil.service for awaiting up the network. 

Tail [boot.log](https://github.com/yggdrasil-network/yggdrasil-go/files/6913439/boot.log)
```
авг 01 11:00:25 fallenedem systemd[1]: Starting yggdrasil...
авг 01 11:00:25 fallenedem systemd[1]: Started yggdrasil.
авг 01 11:00:25 fallenedem yggdrasil[984]: 2021/08/01 11:00:25 WARNING: Tunnel routing is no longer supported
авг 01 11:00:25 fallenedem yggdrasil[984]: 2021/08/01 11:00:25 WARNING: The "SigningPrivateKey" configuration option has been renamed to "PrivateKey"
авг 01 11:00:25 fallenedem yggdrasil[984]: 2021/08/01 11:00:25 Build name: yggdrasil
авг 01 11:00:25 fallenedem yggdrasil[984]: 2021/08/01 11:00:25 Build version: 0.4.0
авг 01 11:00:25 fallenedem yggdrasil[984]: 2021/08/01 11:00:25 Starting up...
авг 01 11:00:25 fallenedem yggdrasil[984]: 2021/08/01 11:00:25 Failed to start TCP interface
авг 01 11:00:25 fallenedem yggdrasil[984]: 2021/08/01 11:00:25 Failed to start link interfaces
авг 01 11:00:25 fallenedem yggdrasil[984]: 2021/08/01 11:00:25 An error occurred during startup
авг 01 11:00:25 fallenedem yggdrasil[984]: 2021/08/01 11:00:25 Listening for TCP on: 127.0.0.1:5003
авг 01 11:00:25 fallenedem yggdrasil[984]: panic: listen tcp 192.168.3.100:5003: bind: cannot assign requested address
авг 01 11:00:25 fallenedem yggdrasil[984]: goroutine 6 [running]:
авг 01 11:00:25 fallenedem yggdrasil[984]: main.run(0x0, 0x7ffe8fc5cee0, 0x13, 0x0, 0x55fe4922ca18, 0x6, 0x0, 0x55fe4922c3aa, 0x4, 0x55fe49326fd8, ...)
[...]

PR #841, opened 2021-09-22
network.target → network-online.target

Fhiss:
Service starts after the network is really on-line. Otherwise, in some cases, the firewall may not see TUN Yggdrasil.

### DIFF
--- a/contrib/systemd/yggdrasil.service
+++ b/contrib/systemd/yggdrasil.service
@@ -1,8 +1,8 @@
 [Unit]
 Description=yggdrasil
-Wants=network.target
+Wants=network-online.target
 Wants=yggdrasil-default-config.service
-After=network.target
+After=network-online.target
 After=yggdrasil-default-config.service
 
 [Service]


neilalexander:
This is a duplicate of #826 and we closed that too. The problem is specific to older versions of Ubuntu seemingly (and maybe others based on it, since it seems distribution maintainers can't leave good enough alone) and has been fixed in more recent releases.

Fhiss:
I have a Fedora 34. The problem is reproduced as follows:
1. Add a WireGuard to the NetworkManager.
2. Connect to the Wi-Fi network where the UDP is blocked.
3. Turn WireGuard off.
4. **journalctl -fu yggdrasil.service**
	_Not multicasting on wlan0 due to error: listen tcp [fe80:db8::%wlan0]:0: bind: cannot assign requested address
	Connected TLS_
5. **firewall-cmd --get-active-zones**
_public
  interfaces: wlan0_

PR #861, opened 2021-11-02
Fix panic in `address.GetKey()`

### LINKED ISSUE #860: crash when accessing some addresses
If you try to ping some very high address like `2ff::` yggdrasil crashes. I can't provide much more info, you can try for yourself, it is very easy to reproduce the bug.

neilalexander:
This should fix #860. There might be a better fix for this, but apparently we overflow `key` and panic in some cases.


Arceliar (approved):
The key/address conversion code could probably use a rewrite at some point, but I think this is OK for now.

I think this bug has been around since before the initial release. Kind of amazed it stuck around this long before causing problems.

Issue #863, opened 2021-11-04
Android build failed for v0.4.2


vikulin:
Android build failed for 0.4.2:
```
vadym@ubuntu:~/yggdrasil-go$ ANDROID=true ./build
Building aar for Android
go: downloading golang.org/x/mobile v0.0.0-20211103151657-e68c98865fb2
go: downloading golang.org/x/tools v0.1.8-0.20211022200916-316ba0b74098
go: downloading golang.org/x/sys v0.0.0-20210809222454-d867a43fc93e
go get: added golang.org/x/mobile v0.0.0-20211103151657-e68c98865fb2
go get: upgraded golang.org/x/net v0.0.0-20210610132358-84b48f89b13b => v0.0.0-20210805182204-aaa1db679c0d
go get: upgraded golang.org/x/sys v0.0.0-20210611083646-a4fc73990273 => v0.0.0-20210809222454-d867a43fc93e
gomobile: /usr/bin/gobind -lang=go,java -outdir=/tmp/gomobile-work-560955579 -tags=mobile github.com/yggdrasil-network/yggdrasil-extras/src/mobile github.com/yggdrasil-network/yggdrasil-go/src/config failed: exit status 1
no exported names in the package "github.com/yggdrasil-network/yggdrasil-extras/src/mobile"
no exported names in the package "github.com/yggdrasil-network/yggdrasil-extras/src/mobile"
no exported names in the package "github.com/yggdrasil-network/yggdrasil-extras/src/mobile"
no exported names in the package "github.com/yggdrasil-network/yggdrasil-extras/src/mobile"
vadym@ubuntu:~/yggdrasil-go$
```

neilalexander:
`go get github.com/yggdrasil-network/yggdrasil-extras@21d029d`

I am not decided yet whether we should move `src/mobile` into the main repository or just remove `ANDROID`/`IOS` support from the build script. It is a nightmare having this circular `go.mod` dependency.

Issue #823, opened 2021-07-30
Error when trying to build for android


jdelrue:
jdelrue@jdelruexps:~/jimbercode/yggdrasil-go$ ./build -a
Building aar for Android
gomobile: /home/jdelrue/go/bin/gobind -lang=go,java -outdir=/tmp/gomobile-work-923186444 -tags=mobile github.com/yggdrasil-network/yggdrasil-extras/src/mobile github.com/yggdrasil-network/yggdrasil-go/src/config failed: exit status 1
no exported names in the package "github.com/yggdrasil-network/yggdrasil-extras/src/mobile"
no exported names in the package "github.com/yggdrasil-network/yggdrasil-extras/src/mobile"
no exported names in the package "github.com/yggdrasil-network/yggdrasil-extras/src/mobile"
no exported names in the package "github.com/yggdrasil-network/yggdrasil-extras/src/mobile"

jdelrue:
Fixed after fiddling around for a few hours. Apperently you need the future branch of yggdrasil-extras, among some specific go requirement. For the ones intrested I created a Docker that builds the android aar:

https://github.com/jdelrue/yggdrasil-mobile-docker

PR #864, opened 2021-11-04
Move `src/mobile` into main repository

neilalexander:
It is a pain to have to maintain `yggdrasil-extras` separately so it seems like it is probably logical at this point to fold this code into the main repository now that it is only a single package and not several. Then I can delete `yggdrasil-extras` altogether.

If we want to `./build` from here then we need `go.mod` to constantly pin the right version of `yggdrasil-extras`, but `yggdrasil-extras` also needs to pin the latest version of Ygg, so one or other is always out of date.

In addition, if we want to stamp the build with the version number, we need the history from this repository anyway. Calling `gomobile` on `yggdrasil-extras` directly doesn't have the correct `LDFLAGS` otherwise.

Apparently the way it is today is a common gotcha — #863, #823.


Arceliar (changes requested):
To my mind, the obviously correct thing to do is to keep this in a separate package, give that package its own `build` script, and get rid of the `gomobile` support in the main repo. I really don't want mobile support in the main repo -- this is an experiment, not a product we're trying to ship, so this isn't something I'm willing to provide support for / help debug. Living under `src/mobile` would tend to indicate that it's supported (at least on a best effort basis), and I'd rather not commit to that.

We already have a directory for unsupported code that we allow in the main repo -- `contrib`. So if this is moved to `contrib/mobile` instead, then I'd (very reluctantly) approve this for now. Ideally, I would prefer to still give it its own build script (under that directory) so we can clean up the main `./build` script, so it's easier to move this into its own package if/when we figure out a better way to do the version stuff, but I suppose the separate build script and cleanup could come later.

I see mobile bindings as being (approximately) synonymous with packaging, and `contrib` is full of package-related scripts already, so I think this would be reasonably consistent.

neilalexander:
Happy with this approach. I've moved it into `contrib` and separated out the build script, tidying up the top-level one.

PR #888, opened 2022-01-30
Restore `uptime`, `bytes_sent` and `bytes_recvd` to `getPeers`

neilalexander:
These were removed at v0.4 because Ironwood didn't provide us with the information. Now we will wrap the conn so we can track them instead.


Arceliar (approved):
Looks good to me.

Issue #886, opened 2022-01-17
question: DNS lookup in isolated network


fps:
Hi,

I'm aware that yggdrasil is aimed at transport/routing and name resolution is out of scope. I just wonder how other people solve name lookup in their networks? So this is strictly aimed at local name resolution (local in the sense that only nodes under my control need to be able to lookup each other - and I also try to keep my network isolated from the main net.)

The simplest would be to maintain a hosts file on all nodes which is highly annoying :)

It would also be possible, I guess, to just have a normal DNS like bind run on a privileged node and have all other nodes use that DNS. That would make life a little bit easier on the other nodes as you would have the names managed centrally. But still a little bit of work which I'm not ready to undertake yet.

Is there a simpler solution? I tried avahi-daemon and explicitly telling it to use tun0 as well, but it doesn't seem to want to play along - it basically just ignores the tun0 interface. And I'm not sure it would work at all anyways since it required multicast.

cofob:
You can try [Alfis](https://github.com/Revertron/Alfis) (small memory and disk footprint) or [EmerDNS](https://emercoin.com/en/emerdns) (large blockchain but simplier to add a new domain, largely supported, for example, on opennic servers).

fps:
Hi! Thanks for the suggestions! Both of these try to solve a larger problem though.

I'm in the process of thinking up a simpler, purely local (local wrt. nodes under one's control) solution.

I'll leave this issue open for now. Maybe there is another solution.

fps:
I played around with ygdrasil a little and came up with this: https://github.com/fps/yggdrasil-go/tree/derived-keys

In that branch of that fork if two nodes share the same private key (and have MixinHostname set to true in their config), they can lookup each other's IP address like this:

on host_a:

<pre>
sudo ./yggdrasil -useconffile conf -hostname  host_b -address
201:d90e:e8ae:9095:1774:c69a:880e:a8ec
</pre>

on host_b:

<pre>
sudo ./yggdrasil -useconffile conf -hostname  host_a -address
200:8402:6332:d1e0:e5f7:dfaa:da5a:50f0
</pre>

(actually since both use the same private key, it doesn't matter on what host the lookup is done)..

marek22k:
You could also try Meshname (https://github.com/zhoreeq/meshname/). Namecoin is something like Emercoin, but due abuse Opennic doesn't support it anymore.

fps:
[...]

PR #894, opened 2022-02-03
limited mDNS support

fps:
I took the idea from my last comment in this thread...

https://github.com/yggdrasil-network/yggdrasil-go/issues/886

...and ran with it a little more. This PR is not a serious pull request, as the code quality is abysmal :) but rather a proof of concept.

With this branch you can setup several nodes with the identical PrivateKey/PublicKey pair in the config and each node will derive a new private key by mixing in the hostname. This in turn makes the public keys and therefore the IP addresses predictable for all nodes sharing the same "master" key.

I have added code to listen for mDNS packets going to the multicast address ff02::fb on the tun interface and see if they contain DNS queries in the form hostname.ygg.local. If they do a response is generated using the above mixing mechanism.

<pre>
[fps@plume:~/src/projects/yggdrasil-fps]$ hostname
plume
</pre>

Resolving plume's IP address:

<pre>
[fps@plume:~/src/projects/yggdrasil-fps]$ avahi-resolve --name plume.ygg.local
plume.ygg.local 202:ed6c:3636:5037:f6d9:6e18:44ff:1f8
</pre>

Resolving another machine's IP address:

<pre>
[fps@plume:~/src/projects/yggdrasil-fps]$ avahi-resolve --name nixos.ygg.local
nixos.ygg.local 202:e2dd:4c4c:e853:96f5:dd34:569f:a79b
</pre>

And finally logging in:

<pre>
[fps@plume:~/src/projects/yggdrasil-fps]$ ssh nixos.ygg.local
ssh: Could not resolve hostname nixos.ygg.local: Device or resource busy

[fps@plume:~/src/projects/yggdrasil-fps]$ ssh nixos.ygg.local
ssh: Could not resolve hostname nixos.ygg.local: Device or resource busy

[fps@plume:~/src/projects/yggdrasil-fps]$ ssh nixos.ygg.local
(fps@nixos.ygg.local) Password: 
Last login: Thu Feb  3 21:42:01 2022 from 202:ed6c:3636:5037:f6d9:6e18:44ff:1f8
</pre>

It's not perfect. But before spending any more energy on it I'd like to discuss the approach.

A note about windows: the mDNS code in windows 10 does not resolve .local addresses with subdomains like .ygg.local. But since the approach in this PR would respond to every hostname thrown at it, regardless whether there really exists a node with that hostname we would need another way to filter requests, e.g. with a prefix like <code>ygg-</code>. So a hostname would be e.g. <code>ygg-plume.local</code>.

A note about avahi: You need to allow avahi to use point-to-point interfaces like tun0. The option is called allowPointToPoint. Enable that, otherwise avahi will ignore the tun interface.

### DIFF
--- a/build
+++ b/build
@@ -28,7 +28,7 @@ if [ -z $TABLES ] && [ -z $DEBUG ]; then
   LDFLAGS="$LDFLAGS -s -w"
 fi
 
-for CMD in yggdrasil yggdrasilctl ; do
+for CMD in yggdrasil yggdrasilctl yggmdns ; do
   echo "Building: $CMD"
   go build $ARGS -ldflags="$LDFLAGS" -gcflags="$GCFLAGS" ./cmd/$CMD
 
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -32,6 +32,7 @@ import (
 	"github.com/yggdrasil-network/yggdrasil-go/src/ipv6rwc"
 	"github.com/yggdrasil-network/yggdrasil-go/src/multicast"
 	"github.com/yggdrasil-network/yggdrasil-go/src/tuntap"
+	"github.com/yggdrasil-network/yggdrasil-go/src/util"
 	"github.com/yggdrasil-network/yggdrasil-go/src/version"
 )
 
@@ -194,6 +195,7 @@ type yggArgs struct {
 	useconffile   string
 	logto         string
 	loglevel      string
+	hostname      string
 }
 
 func getArgs() yggArgs {
@@ -208,6 +210,12 @@ func getArgs() yggArgs {
 	getaddr := flag.Bool("address", false, "returns the IPv6 address as derived from the supplied configuration")
 	getsnet := flag.Bool("subnet", false, "returns the IPv6 subnet as derived from the supplied configuration")
 	loglevel := flag.String("loglevel", "info", "loglevel to enable")
+
+	h, err := os.Hostname()
+	if err != nil {
+		h = ""
+	}
+	hostname := flag.String("hostname", h, "hostname (used for mDNS)")
 	flag.Parse()
 	return yggArgs{
 		genconf:       *genconf,
@@ -221,6 +229,7 @@ func getArgs() yggArgs {
 		getaddr:       *getaddr,
 		getsnet:       *getsnet,
 		loglevel:      *loglevel,
+		hostname:      *hostname,
 	}
 }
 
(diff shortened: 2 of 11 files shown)


Arceliar:
This PR is unlikely to be accepted on its own, simply because adding mDNS/hostname support isn't really a priority for the project.

However: some of the machinery may be useful for other reasons. We could, for example, resolve `<hex-encoded-key>.yggkey.local` (or something like `<hex-encoded-key>-yggkey.local`, to avoid windows `.local` / subdomain issues) and have it spit out the IPv6 address associated with the key. We could maybe even treat it like getting a response from an IP->key lookup, so we set up a session attached to *that* key, on the off chance that someone brute forces a collision (mainly for `/64` subnets).

The use case would be for applications that strictly require a domain name to work. This would give us a name they can use, but without needing any DNS-related infrastructure. It's something we've talked about off and on for a couple of years, but I've just never gotten around to adding support for it.

Then also supporting user-configured hostnames (in some format) isn't *that* much additional logic to throw at the problem, since all of the heavy lifting work (for parsing and responding to mDNS packets) is already in place. The implementation for that part may need some work (I haven't reviewed it), but if it's clean enough then I'm not strictly opposed to keeping it.

fps:
Hi!

Thanks for the response. The code is definitely not in a mergable state. I'll work on it a little more and also implement the lookup you mentioned.

fps:
OK, I think I got it to work somewhat okayish. yggmdns now also seems to run and work on windows..  The base32 encoding of public keys was changed to not include padding chars:

<pre>
C:\Windows\System32>ping 5G3KOEZTSIOCJEISNEJENLXBAOLPM5BNOPQRHFKYOQXQFYDHSBVA-yggk.local

Pinging 5G3KOEZTSIOCJEISNEJENLXBAOLPM5BNOPQRHFKYOQXQFYDHSBVA-yggk.local [200:2c92:b1d9:98db:c7b6:dddb:2ddb:72a2] with 32 bytes of data:
Reply from 200:2c92:b1d9:98db:c7b6:dddb:2ddb:72a2: time=116ms
Reply from 200:2c92:b1d9:98db:c7b6:dddb:2ddb:72a2: time=61ms
</pre>

I kind of like this implementation a little better than the one I had previously, where I directly sniffed off tun0.

fps:
Actually I think this does not even have to go into yggdrasil itself at all. It could just as well be an external package. Closing this.

fps:
Just to finish this up. I moved the code here if anyone is interested in it:

https://github.com/fps/yggdrasil-mdns

PR #897, opened 2022-02-09
Setup refactoring

neilalexander:
This PR adds a new `setup` package which is responsible for parsing command line arguments, reading config from `stdin` or the filesystem, setting log levels and running a core standalone Yggdrasil node (with multicast and the admin socket). 

This PR also shuffles around the `config` and `defaults` packages so that they are not circularly dependent in this new model and so that functions for generating configuration actually live in the `config` package now.

Finally, it updates `cmd/yggdrasil` to use this new arrangement.

The idea here is that there's an awful lot of code that has to be duplicated in `cmd/yggdrasil` today if you just want to start up an Yggdrasil node similar to how we do. This makes it much easier to start a node now without duplicating so much code.


NikitaPuzyryov:
This appears to break `yggdrasil -genconf` command. It outputs array of decimal bytes instead of actual configuration.

src/defaults/defaults.go
```diff
@@ -15,30 +18,10 @@ type platformDefaultParameters struct {
 	DefaultConfigFile string
 
 	// Multicast interfaces
-	DefaultMulticastInterfaces []MulticastInterfaceConfig
+	DefaultMulticastInterfaces []string
```
Arceliar, line 21: It looks like `MulticastInterfaceConfig` is no longer being used? I think that would break existing configs and prevent nodes from being able to configure some options that are available in the current setup (separately controlling if the node listens and/or beacons on an interface, and controlling which port it listens on).

Arceliar (changes requested):
It looks like this no longer uses `MulticastInterfaceConfig`, which is needed to configure some options.

src/defaults/defaults.go
neilalexander: Ah, oops, you're right. I'll fix this in a bit.

Arceliar:
@neilalexander I *think* this was (mostly) made obsolete by #929, so can we close this PR? (Or update it in light of those changes, if we still wanted some of the refactoring)

PR #900, opened 2022-02-22
add command to convert pubkey to ipv6 addr

vdloo:
standalone cli tool like genkeys/main.go that uses
the address.AddrForKey function to convert a pubkey
to an IPv6 address.

```
$ go run main.go <thepubkey>.
IP: <theipv6ip>
```


neilalexander:
Thanks for the PR but we already support getting an address from a config with `-useconffile ... -address` so not sure we need extra utilities floating around.

PR #902, opened 2022-03-18
Fix printing self coordinates in getself command of yggdrasilctl

Rubikoid:
Very simple fix, subj self-explainable.


Jorropo (approved):
Works

PR #907, opened 2022-04-03
Fix regression in Path MTU discovery

neilalexander:
In the past we used to send back anything up to 900 bytes of the packet in the ICMPv6 Packet Too Big response, whereas now we seemingly only send back 40 bytes.

It turns out that sending back only the 40 bytes of IPv6 headers isn't enough for most operating systems to positively ID the flow to reduce the MTU. This PR updates it so that we can send up to 512 bytes instead (900 is probably excessive) — that should leave plenty of room for any number of IPv6 extension headers and the next protocol headers and some of the payload.

This seems to fix the problem in my testing.


Arceliar (approved):
Nice catch. This looks good to me.

PR #909, opened 2022-04-11
Build fails on go 1.16, update docs to say 1.17

gmorenz:
Trying to build with go 1.16 fails with the error

```
Building: yggdrasil
../go/pkg/mod/golang.zx2c4.com/wireguard@v0.0.0-20211017052713-f87e87af0d9a/tun/tun_linux.go:24:2: //go:build comment without // +build comment
src/tuntap/tun.go:20:2: //go:build comment without // +build comment
```

Upgrading to go 1.17 fixes this.

I'm guessing that the 1.16 version in the readme is documentation of what works more than a target, if that's wrong sorry for not filing as an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ or tools in the `contrib` folder.
 If you want to build from source, as opposed to installing one of the pre-built
 packages:
 
-1. Install [Go](https://golang.org) (requires Go 1.16 or later)
+1. Install [Go](https://golang.org) (requires Go 1.17 or later)
 2. Clone this repository
 2. Run `./build`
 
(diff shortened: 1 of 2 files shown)


cathugger:
maybe also should update go.mod so that it fails with understandable error instead of obscure one

gmorenz:
Good point, I'm not super familiar with go tooling and overlooked that.

It (go build) asked me to run `go mod tidy` after changing the version, that added some indirect dependencies to go.mod for some reason? It seems unlikely that that's an issue, but I thought it should call it out in case it is.

cathugger:
I believe they changed the way some tests-related dependencies are calculated, and that shouldn't be an issue because of lazy module loading i think.

neilalexander:
This is done on `develop`.

PR #911, opened 2022-04-18
Version 0.4.4

neilalexander:
### Fixed

- ICMPv6 "Packet Too Big" payload size has been increased, which should fix Path MTU Discovery (PMTUD) when two nodes have different `IfMTU` values configured
- A crash has been fixed when handling debug packet responses
- `yggdrasilctl getSelf` should now report coordinates correctly again

### Changed

- Go 1.17 is now required to build Yggdrasil


neilalexander:
Builds available to test: https://nightly.link/yggdrasil-network/yggdrasil-go/actions/runs/2184355573

Issue #908, opened 2022-04-05
Logging settings inside yggdrasil.conf


0x4E69676874466F78:
At the moment, this is set by a key `-logto` in the service https://github.com/yggdrasil-network/yggdrasil-go/blob/620b901473dd864b7c0c461b001cc38c36776de8/contrib/msi/build-msi.sh#L163
It's not very convenient. Please move the logging settings inside yggdrasil.conf

beryll1um:
I can implement it. Contributors, can approve a feature?

zhoreeq:
> It's not very convenient. Please move the logging settings inside yggdrasil.conf

Why?

0x4E69676874466F78:
@zhoreeq because if I want to turn off logging (or change path) I have to get into the registry where to change the service.
I expect to see logging settings in the config.
Settings must be set in one place and the service configuration is not the right place (counter question: why not store the yggdrasil settings in the registry then? I'm not suggesting this just for thought).


@beryll1um thanks.

zhoreeq:
@0x4E69676874466F78 It works perfectly fine for Linux, no reason to change it. 

Windows users should not manually edit config files, but use a GUI application wrapper for convenience. There are a few at github already.

beryll1um:
> @0x4E69676874466F78 It works perfectly fine for Linux, no reason to change it. 
> 
> Windows users should not manually edit config files, but use a GUI application wrapper for convenience. There are a few at github already.

Are we making life harder for windows users with this change?  Another reason to do it =)

Seriously, I don't think it's a big problem.
Unless we can do what I was talking about initially. Leaving the option to configure it from arguments, which doesn't seem like a very good solution to me.

andrius4669:
aaaaaa please don't

neilalexander:
For now we'll leave this as-is so we can control the log level whilst using `-autoconf`.

PR #913, opened 2022-04-19
Move `-logto` and `-loglevel` parameters to configuration file

beryll1um:
Issue: #908


neilalexander:
Thanks for the PR but closing as we would lose the ability to control it with `-autoconf` otherwise.

PR #924, opened 2022-07-03
Added support for the second level of encryption by a symmetric algorithm (AES).

ufm:
To use it, you need to add a section in the config
Secrets { "public key": "Password"}
You can specify "all" as a publickey - then encryption will be the default for all addresses.
Password should be 16, 24 or 32 character.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/yggdrasil-network/yggdrasil-go
 go 1.17
 
 require (
-	github.com/Arceliar/ironwood v0.0.0-20211125050254-8951369625d0
+	github.com/Arceliar/ironwood v0.0.0-20220409035209-b7f71f05435a
 	github.com/Arceliar/phony v0.0.0-20210209235338-dde1a8dca979
 	github.com/cheggaaa/pb/v3 v3.0.8
 	github.com/gologme/log v1.2.0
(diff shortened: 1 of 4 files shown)


ufm:
> this would mark this traffic as distinct, it is unwise to do so imo. better to have it generate the shared secret with that parameter as an optional additional salt.

When I made this patch I wanted to get:
1.The easiest way to create a "VPN". Now, for this it is enough to write three lines in the config on all VPN nodes:
Secrets: {
"all": "1234567890123456"
}
After that, these nodes will only be able to communicate with each other.

2. Don't break existing functionality.
Now these nodes will still be able to transmit transit traffic. And other nodes that do not use this patch will also be able to transmit traffic from nodes using VPN. At least as long as typeSessionEncTraffic does not intersect with anything. By the way, @neilalexander , I would really appreciate if this type were reserved for this patch, even if the patch itself is not accepted.

3. The simplest patch.
Knowing that a patch might never be accepted, I wanted to "break" existing sources as little as possible.

As a result, this may not be the best way to solve the problem, but I got everything I wanted to get.

majestrate:
you dont need to modify yggdrasil to have such functionality. the easier and more manageable way is with an os firewall with a managed ip whitelist.

majestrate:
you can have the same net effect here without adding unneeded new message types by having the kdf used by each side require an optional out of band shared secret. i dont think you need to add a new message type, given the existing encryption is authenticated.

ufm:
Yes, you are right. It is possible to send encrypted packets with the typeSessionTraffic type, this should not break anything.
This is just a protection for normal hosts - they drop a packet with a type they don't know. I repeat, one of the tasks was - "do not break anything."

neilalexander:
Interesting proof-of-concept but this has a support overhead I don't think we want to take on at this stage.

PR #927, opened 2022-07-11
argument to change uid/gid

cathugger:
different from https://github.com/yggdrasil-network/yggdrasil-go/pull/817 in that it can resolve user names, automatically use user's primary gid & allows specifying gid in the same argument, with `:` eg `username:groupname`.
feel free to criticize & suggest different argument name & description because i didn't put much of thought to that.

### DIFF
--- a/cmd/yggdrasil/chuser_unix.go
+++ b/cmd/yggdrasil/chuser_unix.go
@@ -1,0 +1,73 @@
+//go:build aix || darwin || dragonfly || freebsd || linux || netbsd || openbsd || solaris
+// +build aix darwin dragonfly freebsd linux netbsd openbsd solaris
+
+package main
+
+import (
+	"fmt"
+	osuser "os/user"
+	"strconv"
+	"strings"
+	"syscall"
+)
+
+func chuser(user string) error {
+	group := ""
+	if i := strings.IndexByte(user, ':'); i >= 0 {
+		user, group = user[:i], user[i+1:]
+	}
+
+	u := (*osuser.User)(nil)
+	g := (*osuser.Group)(nil)
+
+	if user != "" {
+		if _, err := strconv.ParseUint(user, 10, 32); err == nil {
+			u, err = osuser.LookupId(user)
+			if err != nil {
+				return fmt.Errorf("failed to lookup user by id %q: %v", user, err)
+			}
+		} else {
+			u, err = osuser.Lookup(user)
+			if err != nil {
+				return fmt.Errorf("failed to lookup user by name %q: %v", user, err)
+			}
+		}
+	}
+	if group != "" {
+		if _, err := strconv.ParseUint(group, 10, 32); err == nil {
+			g, err = osuser.LookupGroupId(group)
+			if err != nil {
+				return fmt.Errorf("failed to lookup group by id %q: %v", user, err)
+			}
+		} else {
+			g, err = osuser.LookupGroup(group)
+			if err != nil {
+				return fmt.Errorf("failed to lookup group by name %q: %v", user, err)
+			}
+		}
+	}
+
+	if g != nil {
+		gid, _ := strconv.ParseUint(g.Gid, 10, 32)
+		err := syscall.Setgid(int(gid))
+		if err != nil {
+			return fmt.Errorf("failed to setgid %d: %v", gid, err)
+		}
+	} else if u != nil {
+		gid, _ := strconv.ParseUint(u.Gid, 10, 32)
+		err := syscall.Setgid(int(gid))
+		if err != nil {
+			return fmt.Errorf("failed to setgid %d: %v", gid, err)
+		}
+	}
+
+	if u != nil {
+		uid, _ := strconv.ParseUint(u.Uid, 10, 32)
+		err := syscall.Setuid(int(uid))
+		if err != nil {
+			return fmt.Errorf("failed to setuid %d: %v", uid, err)
+		}
+	}
+
+	return nil
+}
--- a/cmd/yggdrasil/chuser_unix.go
+++ b/cmd/yggdrasil/chuser_unix.go
@@ -4,7 +4,9 @@
 package main
 
 import (
+	"errors"
 	"fmt"
+	"math"
 	osuser "os/user"
 	"strconv"
 	"strings"
@@ -51,11 +53,11 @@ func chuser(user string) error {
 		gid, _ := strconv.ParseUint(g.Gid, 10, 32)
 		var err error
 		if gid < math.MaxInt {
-				err = syscall.Setgid(int(gid))
+			err = syscall.Setgid(int(gid))
 		} else {
-				err = errors.New("gid too big")
+			err = errors.New("gid too big")
 		}
-		
+
 		if err != nil {
 			return fmt.Errorf("failed to setgid %d: %v", gid, err)
 		}
@@ -71,11 +73,11 @@ func chuser(user string) error {
 		uid, _ := strconv.ParseUint(u.Uid, 10, 32)
 		var err error
 		if uid < math.MaxInt {
-				err = syscall.Setuid(int(uid))
+			err = syscall.Setuid(int(uid))
 		} else {
-				err = errors.New("uid too big")
+			err = errors.New("uid too big")
 		}
-		
+
 		if err != nil {
 			return fmt.Errorf("failed to setuid %d: %v", uid, err)
 		}
--- a/cmd/yggdrasil/chuser_unix.go
+++ b/cmd/yggdrasil/chuser_unix.go
@@ -69,7 +69,13 @@ func chuser(user string) error {
 
 	if u != nil {
 		uid, _ := strconv.ParseUint(u.Uid, 10, 32)
-		err := syscall.Setuid(int(uint32(uid)))
+		var err error
+		if uid < math.MaxInt {
+				err = syscall.Setuid(int(uid))
+		} else {
+				err = errors.New("uid too big")
+		}
+		
 		if err != nil {
 			return fmt.Errorf("failed to setuid %d: %v", uid, err)
 		}
--- a/cmd/yggdrasil/chuser_unix.go
+++ b/cmd/yggdrasil/chuser_unix.go
@@ -49,7 +49,13 @@ func chuser(user string) error {
 
 	if g != nil {
 		gid, _ := strconv.ParseUint(g.Gid, 10, 32)
-		err := syscall.Setgid(int(uint32(gid)))
+		var err error
+		if gid < math.MaxInt {
+				err = syscall.Setgid(int(gid))
+		} else {
+				err = errors.New("gid too big")
+		}
+		
 		if err != nil {
 			return fmt.Errorf("failed to setgid %d: %v", gid, err)
 		}
--- a/cmd/yggdrasil/chuser_unix.go
+++ b/cmd/yggdrasil/chuser_unix.go
@@ -49,13 +49,13 @@ func chuser(user string) error {
 
 	if g != nil {
 		gid, _ := strconv.ParseUint(g.Gid, 10, 32)
-		err := syscall.Setgid(int(gid))
+		err := syscall.Setgid(int(uint32(gid)))
 		if err != nil {
 			return fmt.Errorf("failed to setgid %d: %v", gid, err)
 		}
 	} else if u != nil {
 		gid, _ := strconv.ParseUint(u.Gid, 10, 32)
-		err := syscall.Setgid(int(gid))
+		err := syscall.Setgid(int(uint32(gid)))
 		if err != nil {
 			return fmt.Errorf("failed to setgid %d: %v", gid, err)
 		}
@@ -63,7 +63,7 @@ func chuser(user string) error {
 
 	if u != nil {
 		uid, _ := strconv.ParseUint(u.Uid, 10, 32)
-		err := syscall.Setuid(int(uid))
+		err := syscall.Setuid(int(uint32(uid)))
 		if err != nil {
 			return fmt.Errorf("failed to setuid %d: %v", uid, err)
 		}
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -52,6 +52,7 @@ func main() {
 	getsnet := flag.Bool("subnet", false, "use in combination with either -useconf or -useconffile, outputs your IPv6 subnet")
 	getpkey := flag.Bool("publickey", false, "use in combination with either -useconf or -useconffile, outputs your public key")
 	loglevel := flag.String("loglevel", "info", "loglevel to enable")
+	chuserto := flag.String("user", "", "user (and, optionally, group) to set UID/GID to")
 	flag.Parse()
 	
 	done := make(chan struct{})
@@ -279,6 +280,14 @@ func main() {
 		// Wait for all parts to shutdown properly
 		<-done
 	})
+
+	// Change user if requested
+	if *chuserto != "" {
+		err = chuser(*chuserto)
+		if err != nil {
+			panic(err)
+		}
+	}
 
 	// Block until we are told to shut down.
 	<-ctx.Done()
--- a/cmd/yggdrasil/chuser_other.go
+++ b/cmd/yggdrasil/chuser_other.go
@@ -1,0 +1,10 @@
+//go:build !aix && !darwin && !dragonfly && !freebsd && !linux && !netbsd && !openbsd && !solaris
+// +build !aix,!darwin,!dragonfly,!freebsd,!linux,!netbsd,!openbsd,!solaris
+
+package main
+
+import "errors"
+
+func chuser(user string) error {
+	return errors.New("setting uid/gid is not supported on this platform")
+}


cmd/yggdrasil/main.go
```diff
@@ -361,6 +364,13 @@ func run(args yggArgs, ctx context.Context, done chan struct{}) {
 		logger.Errorln("An error occurred starting TUN/TAP:", err)
 	}
 	n.tuntap.SetupAdminHandlers(n.admin)
+	// Change user if requested
+	if args.chuser != "" {
```
Arceliar, line 368: It may make sense to change to `nobody:nobody` by default, so things are a little less unsafe for people who just `sudo yggdrasil -autoconf` or similar. Thoughts @neilalexander?

rany2: @Arceliar I think it should be noted that `nobody:nobody` is not standard across Linux distros. For example, Debian and friends use `nobody:nogroup` instead. We should check which group is available first or maybe use something like Dynamic Users in systemd (use a random UID/GID that's currently not in use by the system and reserve it).

cathugger: could just use `nobody` and maybe it would just resolve primary group on its own i think?
also would need to check UID being 0 before doing chuser by default (as this can be started as own user if right caps are given).
and opting out may be kinda uglier - manually specifying empty user string, though i can't really find many reasons of opting out, other than some possible exotic distributions/installations without proper user databases - which i guess i could imagine in some sterilized docker environments or routers but i don't really know how common, and probably checking existence of `nobody` user before deciding to default to chuser'ing to it would make sense - and we could just check database before setting default to `nobody`, likely writing function in chuser_unix.go for that (& stub returning empty string in chuser_other.go).
Requesting opinions for this mindfuzz of mine i guess

cathugger: I thought about this a bit more and probably programming in this kind of logic into daemon itself is too much of a magical behavior imo, and this should be provided in distributions' `.service` files instead.

rany2: well there isn't any need for this to be in any distro, Debian has a pretty hardened default that runs as yggdrasil user from the start. I think all you need to get it to work is CAP_NET_ADMIN, I think this is how Debian configured their OpenVPN services too

cathugger: Non-systemd gentoo and alpine would benefit from this, as from my experience OpenRC scripts can't easily give out capabilities.

cathugger:
I looked around a bit, and git daemon implements this in a bit cleaner way, maybe, command line arguments wise:
```
   --user=<user>, --group=<group>
       Change daemon's uid and gid before entering the service loop. When
       only --user is given without --group, the primary group ID for the
       user is used. The values of the option are given to getpwnam(3) and
       getgrnam(3) and numeric IDs are not supported.

       Giving these options is an error when used with --inetd; use the
       facility of inet daemon to achieve the same before spawning git
       daemon if needed.

       Like many programs that switch user id, the daemon does not reset
       environment variables such as $HOME when it runs git programs, e.g.
       upload-pack and receive-pack. When using this option, you may also
       want to set and export HOME to point at the home directory of
       <user> before starting the daemon, and make sure any Git
       configuration files in that directory are readable by <user>.
```
Though [nsd](https://linux.die.net/man/8/nsd) for example uses different syntax, using `-u` and taking in `username, id, or id.gid`, which is more similar to what I've implemented here.
I think that current implementation is kinda fine, but maybe documentation could be improved, but I can't really think of anything good atm.

cmd/yggdrasil/main.go
cathugger: Oh also *BSDs

cathugger: And [openwrt](https://github.com/openwrt/packages/blob/openwrt-21.02/net/yggdrasil/files/yggdrasil.init)

rany2: It's worth mentioning that on Linux systems without systemd you could simply do: `sudo capsh --user='yggdrasil' --inh='cap_net_admin,cap_net_bind_service' --addamb='cap_net_admin,cap_net_bind_service' --shell=/bin/sh -- -c 'exec yggdrasil -autoconf'`

cathugger: I'm aware of capsh, and tried using it in past for exactly this purpose (which ended looking too much of a mess for my tastes), but evidently many people either don't know, or they don't like such an obscure & hacky looking line... https://git.alpinelinux.org/aports/tree/community/yggdrasil/yggdrasil.initd for example.
I also just tried it myself & it failed with `Admin socket failed to listen: listen unix /var/run/yggdrasil.sock: bind: permission denied` so it's not so simple i guess...

cathugger: Also as far as i can see in [yggdrasil.service](https://github.com/yggdrasil-network/yggdrasil-go/blob/develop/contrib/systemd/yggdrasil.service), it doesn't even specify `User`, so appears to run as root, at least in my archlinux system.

rany2: You should be able to genconf and disable the unix socket/move it to a location yggdrasil user could write to

cathugger: Weird, I can't see creation of `/var/run/yggdrasil` or yggdrasil user (only group) in https://github.com/yggdrasil-network/yggdrasil-go/blob/develop/contrib/deb/generate.sh
Am I looking at wrong place?

rany2: Debian being Debian maintains their own fork of yggdrasil with their own enhancements.

cathugger: I see. This packaging is indeed fairly good. This has gotten a bit offtopic, but it would be cool if one could specify default socket / config path at compile time (eg using `go build -ldflags="-X defaults.defaultAdminListen=unix:///var/run/yggdrasil/yggdrasil.sock"`) so that linux distros wouldn't need patches there to achieve saner packaging.
Either way this repo's systemd files suck at running non-root, and so consequentially do vendor (not official debian's) debs and archlinux package.

neilalexander: On the whole, I'm very pro-having separate flags for UID and GID, because then it is clearer that we can do one without the other, and it is easier for packagers to do the right thing for their distribution/platform.

I'm not entirely sure whether we want to do that *by default* — we could certainly do some testing to see how it feels. The main thing I can think of that might break is setting up new multicast sockets when an interface disappears and reappears later but we've already dropped privileges, but maybe it'll be OK?

neilalexander: Actually isn't that what #817 is doing?

cathugger: #817 can take only numeric UID/GIDs, which is unwieldy.
>because then it is clearer that we can do one without the other

why would you want do one but not another? everywhere i looked around, one either wants to set uid and also gid to primary group of uid, or uid/gid separate, which is also possible. with this implementation, one can also set gid without uid (eg `-user ":groupname"` or `-user ":gid"`), though it may be a little bit unobvious, but why would you want to not drop root?

if you think that instead of using `-user uid:gid` syntax we should do `-uid user -gid group`, i can change to that too, looks like a minimal change to me.

>The main thing I can think of that might break is setting up new multicast sockets when an interface disappears and reappears later but we've already dropped privileges, but maybe it'll be OK?

that's sorta fundamentally incompatible with privilege dropping tbh. on linux, we could probably do some magic with caps and reserve the right to deal with interfaces, but on other unixes we can't. restarting daemon when new interfaces to be managed pop up could probably make sense, but i don't quite feel that's what yggdrasil should do on its own either.

cathugger: unless multicast actually works without requiring root (i somehow confused it with managing interfaces, and using multicast aint that)

cathugger:
rebased.

does this need any further work?
should I change argument format / documentation? current stuff is kinda unclear a bit imo but does anyone else find that to be an issue?
splitting to two arguments could make argument documentation easier too probably.
i *think* multicast stuff should work fine with this unless multicast port is picked to be <=1024.

neilalexander:
I haven't forgotten about this — I'd like to get this merged in some shape or other, but let's wait until the v0.5 branch is a bit cleaner before rebasing there.

neilalexander:
If you're happy to rebase one last time on top of `develop`, let's get this merged for v0.5.

cmd/yggdrasil/chuser_unix.go
```diff
@@ -0,0 +1,73 @@
...
+			g, err = osuser.LookupGroup(group)
+			if err != nil {
+				return fmt.Errorf("failed to lookup group by name %q: %v", user, err)
+			}
+		}
+	}
+
+	if g != nil {
+		gid, _ := strconv.ParseUint(g.Gid, 10, 32)
+		err := syscall.Setgid(int(uint32(gid)))
```
Xephobia, line 52: The CodeQL check failure is because you do not check if `gid` fits in `int` (`int` may be smaller than `uint32` so you need to import math and apply the changes (if i'm not wrong):
```suggestion
		var err error
		if gid < math.MaxInt {
				err = syscall.Setgid(int(gid))
		} else {
				err = errors.New("gid too big")
		}
		
```

cmd/yggdrasil/chuser_unix.go
```diff
@@ -0,0 +1,73 @@
...
+		gid, _ := strconv.ParseUint(u.Gid, 10, 32)
+		err := syscall.Setgid(int(uint32(gid)))
+		if err != nil {
+			return fmt.Errorf("failed to setgid %d: %v", gid, err)
+		}
+	}
+
+	if u != nil {
+		uid, _ := strconv.ParseUint(u.Uid, 10, 32)
+		err := syscall.Setuid(int(uint32(uid)))
```
Xephobia, line 66: same thing for this one:
```suggestion
		var err error
		if uid < math.MaxInt {
				err = syscall.Setuid(int(uid))
		} else {
				err = errors.New("uid too big")
		}
		
```

Xephobia (changes requested):

cmd/yggdrasil/chuser_unix.go
```diff
@@ -0,0 +1,87 @@
...
+	"strconv"
+	"strings"
+	"syscall"
+)
+
+func chuser(user string) error {
+	group := ""
+	if i := strings.IndexByte(user, ':'); i >= 0 {
+		user, group = user[:i], user[i+1:]
+	}
```
klemensn, line 20: Why not just `strings.Split(user, ':')`?

cmd/yggdrasil/chuser_unix.go
```diff
@@ -0,0 +1,87 @@
...
+
+	u := (*osuser.User)(nil)
+	g := (*osuser.Group)(nil)
+
+	if user != "" {
+		if _, err := strconv.ParseUint(user, 10, 32); err == nil {
```
klemensn, line 26: Why parse it yourself?  You can try a name lookup, fallback to UID on error and return if that fails, too, just like getpwnam(3) behaves:
```
if u, err = osuser.Lookup(user); err != nil {
    if u, err = osuser.LookupId(user); if err != nil {
        return err
    }
}
```

cmd/yggdrasil/chuser_unix.go
```diff
@@ -0,0 +1,87 @@
...
+		} else {
+			g, err = osuser.LookupGroup(group)
+			if err != nil {
+				return fmt.Errorf("failed to lookup group by name %q: %v", user, err)
+			}
+		}
+	}
+
+	if g != nil {
+		gid, _ := strconv.ParseUint(g.Gid, 10, 32)
```
klemensn, line 53: Questionable for this API to return the *ID* for a known-to-exist group as string...

cmd/yggdrasil/chuser_unix.go
```diff
@@ -0,0 +1,87 @@
...
+		var err error
+		if gid < math.MaxInt {
```
klemensn, line 55: os/user returns ID strings that were "validated" with strconv.Atoi() which returns `int`, so this should always be true.

cmd/yggdrasil/chuser_unix.go
cathugger: `IndexByte` is probably speediest as `Split` allocates slice and uses string for splitting.
`Cut` would probably be more fitting here API niceness wise, but I'm not sure if really worth changing at this point.

cmd/yggdrasil/chuser_unix.go
cathugger: this PR introduced as replacement for one that only used numeric IDs so I preserved ability to do chown/chgrp without passwords database file being present in system.
`LookupId` depends on /etc/passwd being present which is technically not a necessity.

cathugger: nvm we do LookupId after this anyway so it's a moot point.

cmd/yggdrasil/chuser_unix.go
cathugger: had to dig deep to confirm this but indeed.
also could probably be `<=`.
feel free to make a PR if you care for this as this looks like something code being saner internally not something that'll lead to bugs.

cmd/yggdrasil/chuser_unix.go
klemensn: > `IndexByte` is probably speediest as `Split` allocates slice and uses string for splitting.

This code runs once, it is not a hot path.

> `Cut` would probably be more fitting here API niceness wise, but I'm not sure if really worth changing at this point.

Sure, either way looks cleaner than manual indexing.

cmd/yggdrasil/chuser_unix.go
klemensn: > this PR introduced as replacement for one that only used numeric IDs so I preserved ability to do chown/chgrp without passwords database file being present in system.

Do these systems exist?  Are they relevant?

> `LookupId` depends on /etc/passwd being present which is technically not a necessity.

The next line uses that function, so this did not go as planned.

So you might as well clean it up, imho.

cmd/yggdrasil/chuser_unix.go
klemensn: > feel free to make a PR if you care for this as this looks like something code being saner internally not something that'll lead to bugs.

Either that or a clean(er) `chuser_bsd.go` implementation, should I ever start using this option.

Point being:  Complexity hides bugs, so code ought to be as simple as possible.

To my eye, multiple Lookup*() and Parse*() are not as trivial to reason about as it could be.

cmd/yggdrasil/chuser_unix.go
cathugger: as I've mentioned in following comments, yes, this needs pw database anyways, but i prefer minimizing potentially latency introducing tasks (like opening/closing passwd file) even if they're done at init time, so it's guarded by fast parseuint check. it's a matter of taste, eliminating that goes into category of nitpicking IMO.

cmd/yggdrasil/chuser_unix.go
cathugger: without digging deep into the source of `osuser.LookupId` one can't even know whether these ints were checked so I'm not sure complexity argument is even applicable here, it's more complex to verify whether the code is correct if there is no check.
my gut instinct when API doesn't tell stuff like returned ranges of ints to just check myself, because checking more usually catches more bugs than not checking stuff like returns.
IMO if check is eliminated, then strconv.Atoi() check behavior should be documented in comment, as otherwise it won't end up more simple.

cathugger: also uint32 vs int is potentially hell on its own due to int being 64bit on 64bit systems and strconv.Atoi() ending up verifying different on 32bit vs 64bit systems but honestly systems just don't use ints as large as these for UIDs/GIDs anyways so i can't really care abt this.

cathugger: https://github.com/yggdrasil-network/yggdrasil-go/pull/927/commits/87fa0a8936d554d3f885ee6f7f6d6647c8c75f65 apparently was someone's else's change too but I can't argue one way or another abt it.

cathugger: to satisfy faulty Code scanning / CodeQL warnings out of all the things

cmd/yggdrasil/main.go
```diff
@@ -280,6 +281,14 @@ func main() {
...
 	})
 
+	// Change user if requested
+	if *chuserto != "" {
+		err = chuser(*chuserto)
+		if err != nil {
+			panic(err)
+		}
+	}
+
```
klemensn, line 291: The main goroutine is racing against those creating a tun(4) interface, raw socket for routing and UNIX socket as control interface, i.e. startup can fail now as yggdrasil may drop to an unprivileged user before having completed all privileged tasks.

Just try `yggdrasil -autoconf -user nobody` a few times:

```
# /usr/ports/pobj/yggdrasil-go-0.5.9/fake-amd64/usr/local/bin/yggdrasil -autoconf -user nobody
2024/11/02 20:24:32 Build name: yggdrasil-go
2024/11/02 20:24:32 Build version: 0.5.9
2024/11/02 20:24:32 Your public key is 82ff9b15b30dc6e460f907cb3aecbe664c024c101efa773b3ecdccf1cb72d818
2024/11/02 20:24:32 Your IPv6 address is 200:fa00:c9d4:99e4:7237:3e0d:f069:8a26
2024/11/02 20:24:32 Your IPv6 subnet is 300:fa00:c9d4:99e4::/64
2024/11/02 20:24:32 Interface name: tun0
2024/11/02 20:24:32 Interface IPv6: 200:fa00:c9d4:99e4:7237:3e0d:f069:8a26/7
2024/11/02 20:24:32 Interface MTU: 16384
2024/11/02 20:24:32 Admin socket failed to listen: listen unix /var/run/yggdrasil.sock: bind: permission denied
```

klemensn:
This code fails to unset supplementary groups, i.e. the process retains whatever access it had even after dropping to a different user and group:

```
# id                                                   
uid=0(root) gid=0(wheel) groups=0(wheel), 2(kmem), 3(sys), 4(tty), 5(operator), 20(staff), 31(guest)
# ./yggdrasil -autoconf -logto /dev/null -user nobody &
[1] 4337
# ps -o command,user,group,supgrp -U nobody
COMMAND          USER     GROUP    SUPGRP
./yggdrasil -aut nobody   nobody   wheel,kmem,sys,tty,operator,staff,guest                         
```

Thus the process still has permissions to open, e.g. a `660 root:wheel` file, which is bad.

cmd/yggdrasil/chuser_unix.go
klemensn: > to satisfy faulty Code scanning / CodeQL warnings out of all the things

This looks related over at #1202:
Screenshot 2024-11-04 at 04-05-21 Set groups when dropping privileges to not leak supplementary group access by klemensn · Pull Request #1202 · yggdrasil-network_yggdrasil-go

klemensn: > feel free to make a PR if you care for this as this looks like something code being saner internally not something that'll lead to bugs.

- `-user ''` and `-user ':'` are silently ignored and the process continues without error or change in privilege.
- `-user :group` change only the group, but not the user, contrary to what the usage says:
```
  -user string
    	user (and, optionally, group) to set UID/GID to
```

I did fix these now in #1203.

PR #929, opened 2022-08-06
Node setup refactoring

neilalexander:
This breaks the dependency on the Yggdrasil core needing to take a reference to a config struct. Instead, we pass in the options at startup as variadic parameters and Yggdrasil maintains the internal state itself in a protected fashion.

This means that unexpected mutations of the config struct from outside of Yggdrasil no longer create data races, and it means that we can now start an Yggdrasil node much more expressively, i.e. from `core_test.go`:
```
if nodeA, err = New(skA, ListenAddress("tcp://127.0.0.1:0"), IfName("none")); err != nil {
	t.Fatal(err)
}
```

Responsibility for parsing the configuration JSON is now given to the `cmd/yggdrasil` itself.

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -36,7 +36,7 @@ import (
 )
 
 type node struct {
-	core      core.Core
+	core      *core.Core
 	config    *config.NodeConfig
 	tuntap    *tuntap.TunAdapter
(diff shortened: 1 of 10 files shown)


Arceliar (approved):
Varadic functions in Go are always a little awkward behind the scenes, but I don't think there's a better way to do this. LGTM.

PR #930, opened 2022-08-09
Fix Android multicast crash

Saiv46:
Closes yggdrasil-network/yggdrasil-android#14

### DIFF
--- a/contrib/mobile/mobile.go
+++ b/contrib/mobile/mobile.go
@@ -62,11 +62,11 @@ func (m *Yggdrasil) StartJSON(configjson []byte) error {
 	if len(m.config.MulticastInterfaces) > 0 {
 		if err := m.multicast.Init(&m.core, m.config, logger, nil); err != nil {
 			logger.Errorln("An error occurred initialising multicast:", err)
-			return err
+			return nil
 		}
 		if err := m.multicast.Start(); err != nil {
 			logger.Errorln("An error occurred starting multicast:", err)
-			return err
+			return nil
 		}
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


contrib/mobile/mobile.go
```diff
@@ -62,11 +62,11 @@ func (m *Yggdrasil) StartJSON(configjson []byte) error {
 	if len(m.config.MulticastInterfaces) > 0 {
 		if err := m.multicast.Init(&m.core, m.config, logger, nil); err != nil {
 			logger.Errorln("An error occurred initialising multicast:", err)
-			return err
+			return nil
```
neilalexander, line 65: What exactly is the error that you're dropping to the floor here?

neilalexander: What is the precise error message?

r4sas: Same as in comment on second block here: https://github.com/yggdrasil-network/yggdrasil-android/issues/14#issuecomment-1125834065

> 2022/08/06 17:52:42 An error occurred starting multicast: listen udp6 [::]:9001: protocol not available

Saiv46: > Why would multicast fail to init on your device?

I don't know exactly why on some devices setting up UDPv6 would fail when using multicast address (unicast addresses still working fine, likely manufacturer applied some networking patches for battery optimization).

So instead, mobile version should behave just like desktop one - just print error message and do nothing if multicast fails.

Saiv46: > Why would multicast fail to init on your device?

I don't know exactly why on some devices setting up UDPv6 would fail when using multicast address (unicast addresses still working fine, likely manufacturer applied some networking patches for battery optimization).

So instead, mobile version should behave just like desktop one - just print error message and do nothing if multicast fails.

neilalexander (approved):

PR #938, opened 2022-08-30
Refactor ipv6rwc to interface instead of double struct

Demacr:
IMO it is better to use interface instead of struct which is another struct.

### DIFF
--- a/contrib/mobile/mobile.go
+++ b/contrib/mobile/mobile.go
@@ -26,7 +26,7 @@ import (
 // in Swift therefore we use the "dummy" TUN interface instead.
 type Yggdrasil struct {
 	core      *core.Core
-	iprwc     *ipv6rwc.ReadWriteCloser
+	iprwc     ipv6rwc.ReadWriteCloser
 	config    *config.NodeConfig
 	multicast multicast.Multicast
 	log       MobileLogger
--- a/src/ipv6rwc/ipv6rwc.go
+++ b/src/ipv6rwc/ipv6rwc.go
@@ -298,6 +298,38 @@ func (k *keyStore) writePC(bs []byte) (int, error) {
 
 // Exported API
 
+type ReadWriteCloser interface {
+	Address() address.Address
+	Subnet() address.Subnet
+	MTU() uint64
+	MaxMTU() uint64
+	SetMTU(uint64)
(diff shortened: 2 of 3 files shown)


cathugger:
Why? I don't see any architectural improvements, and this will result in lower performance because of dynamic dispatch.

Demacr:
You are right. But what purpose to use struct ReadWriteCloser which is struct keystore? And use methods each of structs?

Demacr:
By the name `ReadWriteCloser`, like `io.ReadWriteCloser`, I thought someone want to do interface but did struct.

cathugger:
Yeah, yggdrasil's codebase aint all idiomatic golang, but I can't see how adding interface that doesn't give any advantages of interface will help (making interfaces with only single implementations isn't what interfaces are for).
There is a history in golang stdlib of structs with names that are interfaces in different packages when given interface doesn't have all the methods necessary for proper function, eg [ed25519.PrivateKey](https://pkg.go.dev/crypto/ed25519@go1.19#PrivateKey) implementing [crypto.PrivateKey](https://pkg.go.dev/crypto@go1.19#PrivateKey).

Demacr:
I agree with 1st point (about single implementation). About 2nd, IMO it should be interface or single struct, with name which doesn't look like interface name (-er ending and implement 3 more methods than name contains) : )

cathugger:
More examples in stdlib to add to 2nd point: [bufio.Writer](https://pkg.go.dev/bufio@go1.19#Writer), [bufio.Reader](https://pkg.go.dev/bufio@go1.19#Reader), [bufio.ReadWriter](https://pkg.go.dev/bufio@go1.19#ReadWriter), and well honestly i'm pretty bored of digging any more for this at this point.
I agree that this shoudn't be about `ReadWriteCloser`, though, as it's sort-of undescriptive of what this is about, especially package name including `rwc` which refers to that, indicating that sole purpose of this package is implementing IPv6 ReadWriteCloser functionality in whatever way, but which seems to be actually more about mapped IPv6 packet buffering...

neilalexander (commented):
I can't really see a clear benefit to this change and it will probably just result in more indirection. It's better for *receivers* to define interfaces IMO and we already conform to the common ones that people expect (like `io.ReadWriteCloser`).

PR #939, opened 2022-09-02
Admin socket and `yggdrasilctl` refactoring

neilalexander:
This PR refactors much of the API surface to have more sensible response bodies and rewrites most of `yggdrasilctl` too.

Some notable changes:
* Request and response structs now have a clearer shape and are used by both the admin socket and `yggdrasilctl`
* Table rendering in `yggdrasilctl` is now a bit cleaner and produces more human-friendly output overall
* Results from the admin API are now sorted where it makes sense to do so, rather than having a random return order


Arceliar (approved):
Not tested, but looks reasonable by eye. I'm fine with merging now and iterating more (if needed) after it's had some time in `develop`.

PR #941, opened 2022-09-04
Link refactor

neilalexander:
This refactors most of the link code to hopefully make it a bit more clear and idiomatic, as well as factorising a few things and tidying up others. It also adds support for peering over UNIX sockets (i.e. `unix:///path/to/socket.sock`).


Arceliar (approved):
This looks OK to my eyes.

The name `link_unix.go` could cause some confusion, since `unix` is a valid platform in go 1.19, so it could look like a reference to unix platform specific code rather than code for a unix socket... but I'm happy with it as-is for now, unless somebody comes up with a better idea.

PR #949, opened 2022-09-24
Use deadline for link handshake

neilalexander:
This uses a 6 second deadline for timeouts instead of using `util.FuncTimeout` at 30 seconds for the read and then again for the write.

If the handshake doesn't complete within 6 seconds then it's going to probably collapse when we give the connection to Ironwood and it tries to do a keepalive anyway.


Arceliar (approved):
Sounds good to me. We may want to rethink this later, if/when we change how keepalives / timeouts work in ironwood.

PR #951, opened 2022-09-26
Reimplement AddPeer and RemovePeer for admin socket

### LINKED ISSUE #950: addpeer, removepeer missing from admin socket
I'd like to add and remove peers from a running yggdrasil instance. I remember this being a feature but it seems to have been removed?

Basically I want my init system to start a minimally configured yggdrasil and then delegate the further configuration. I can think of a few ways this would be useful:
- I want an address to bind to *as soon as possible* during boot.
- Find peers using DNS service discovery where multicast discovery isn't possible (Freifunk) - without adding the DNS SD logic to yggdrasil-go.
- Peer ~~auto-discovery~~ optimization, I'd like to have the option to dynamically add and remove peers by estimating the physical distance of known public peers. I don't know how I would do that but I can't experiment without being able to externalise that logic.

More configuration options exposed through the control socket is better. It may seem like a feature that people aren't using, but regardless I think it's the state-of-the-art way to do configuration.

ehmry:
Works for me.

Fix #950

### DIFF
--- a/src/core/api.go
+++ b/src/core/api.go
@@ -181,8 +181,10 @@ func (c *Core) SetLogger(log util.Logger) {
 }
 
 // AddPeer adds a peer. This should be specified in the peer URI format, e.g.:
-// 		tcp://a.b.c.d:e
-//		socks://a.b.c.d:e/f.g.h.i:j
+//
+//	tcp://a.b.c.d:e
+//	socks://a.b.c.d:e/f.g.h.i:j
+//
 // This adds the peer to the peer list, so that they will be called again if the
 // connection drops.
 func (c *Core) AddPeer(uri string, sourceInterface string) error {
@@ -190,12 +192,12 @@ func (c *Core) AddPeer(uri string, sourceInterface string) error {
 	if err != nil {
 		return err
 	}
-	err = c.CallPeer(u, sourceInterface)
+	info, err := c.links.call(u, sourceInterface)
 	if err != nil {
 		return err
 	}
 	phony.Block(c, func() {
-		c.config._peers[Peer{uri, sourceInterface}] = struct{}{}
+		c.config._peers[Peer{uri, sourceInterface}] = &info
 	})
 	return nil
 }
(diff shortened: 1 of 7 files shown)


src/core/api.go
```diff
@@ -185,74 +185,24 @@ func (c *Core) SetLogger(log util.Logger) {
...
-		}
-		if _, ok := c.config.Current.InterfacePeers[sintf]; !ok {
-			c.config.Current.InterfacePeers[sintf] = []string{addr}
-		} else {
-			c.config.Current.InterfacePeers[sintf] = append(c.config.Current.InterfacePeers[sintf], addr)
-		}
+	if err := c.CallPeer(u, sourceInterface); err != nil {
+		return err
 	}
+	c.config._peers[Peer{uri, sourceInterface}] = struct{}{}
```
neilalexander, line 196: It isn’t safe to modify `c.config._peers` from outside of the `*Core` actor (hence the underscore at the beginning of the name) — you should surround this with something like:
```
phony.Block(c, func() {
   …
})
```

src/core/api.go
```diff
@@ -185,74 +185,24 @@ func (c *Core) SetLogger(log util.Logger) {
...
-			if addr == peer.intf.name() {
-				c.peers._removePeer(peer)
-			}
-		}
-	})
 
+// RemovePeer removes a peer. The peer should be specified in URI format, see AddPeer.
+// The peer is not disconnected immediately.
+func (c *Core) RemovePeer(uri string, sourceInterface string) error {
+	delete(c.config._peers, Peer{uri, sourceInterface})
```
neilalexander, line 203: Same here.

neilalexander (changes requested):

src/core/api.go
neilalexander: `CallPeer` was right to be outside of the `phony.Block` and to check the error etc as before — only the `c.config._peers` bit needed protecting.

ehmry: Should the error really be checked for `CallPeer`? What if a there is a temporary failure on the first connection?

neilalexander: Well, we can either ignore the fact that the peer might not be reachable in case they come back in the future, or we can make more more immediately apparent to the user that the endpoint was dead. Not sure which is better from an ergonomic perspective.

ehmry: I think its better to return the error immediately and the admin client can just try again later.

neilalexander (approved):

neilalexander:
Thanks for this!

Issue #891, opened 2022-02-01
Incorrect cleanup in service mode


Vort:
When I stop Yggdrasil service, it exits abruptly, without usual "Stopping..." and "Stopped" messages.
This leads to bunch of "orphaned" adapters, which is not good:
![image](https://user-images.githubusercontent.com/1242858/151974081-aaa0280c-24e5-4061-b4ab-0541ca789531.png)
If I start Yggdrasil in console, there is a chance that message like "Removed orphaned adapter "Yggdrasil 11"" appears and one of adapters goes away.
But: 1. It is too inconvenient to restart Yggdrasil over and over in hopes of cleanup. 2. You can't do it manually:
![image](https://user-images.githubusercontent.com/1242858/151974840-171cf519-6058-4f9b-bea6-e30d9127bea8.png)

OS: Windows 7 SP1 x64
Version: 0.4.2-7-g9f5cc0e

Vort:
@neilalexander exactly the same happens with Build name: yggdrasil-neilalexanderv045 / Build version: 0.4.4-41-g6bda57e : when I stop service, process silently goes away without writing single line to log file while leaving trash adapters behind.
What can I do to help with debugging this problem?

Vort:
After applying 4fbdeb4e3fd77dfc2bef8f9784bf14213d5461e4 to my fork (97c20d78d344c6b71b607c7366be0833f56f066d), I see `Stopping...` and `Stopped` messages in logs. Thanks @Revertron.

Issue #953, opened 2022-10-02
Cgo multicast doesn't build on macOS 13


rex4539:
```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/hfs/hfs_format.h:794:2: error: unknown type name 'uuid_string_t'; did you mean 'io_string_t'?
        uuid_string_t   ext_jnl_uuid;
        ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/device/device_types.h:89:33: note: 'io_string_t' declared here
typedef char                    io_string_t[512];
                                ^

...

/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/hfs/hfs_format.h:796:20: error: use of undeclared identifier 'uuid_string_t'
        char            reserved[JIB_RESERVED_SIZE];
                                 ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/hfs/hfs_format.h:787:61: note: expanded from macro 'JIB_RESERVED_SIZE'
#define JIB_RESERVED_SIZE  ((32*sizeof(u_int32_t)) - sizeof(uuid_string_t) - 48)
                                                            ^
```

neilalexander:
You will need to provide more details than that. 

Which version of macOS, which version of Xcode and which version of Go?

rex4539:
macOS 13.0 (22A5358e)
Xcode 14.1 beta 3 (14B5033e)
go version go1.19.1 darwin/amd6

neilalexander:
I'm wondering if this is something to do with the macOS multicast code, can you try building with `CGO_ENABLED=0`?

Failing that, it looks like there's something in the macOS SDK, so would recommend you try with a release version like Xcode 14.0.1.

rex4539:
> I'm wondering if this is something to do with the macOS multicast code, can you try building with `CGO_ENABLED=0`?

This worked.

neilalexander:
OK, glad to hear it. Will keep this issue open and rename it.

neilalexander:
Closing as out-of-date.

rex4539:
Nope. It's still up to date.

neilalexander:
[...]

PR #957, opened 2022-10-15
Version 0.4.5

### LINKED ISSUE #954: [develop] Admin interface does not return user's NodeInfo
Admin interface does not return user's NodeInfo:

Steps:

```
vadym@ubuntu:~/yggdrasil-go$ ./yggdrasilctl getself
Build name:   	yggdrasil-develop                                               	
Build version:	0.4.4-36-g69632ba                                               	
IPv6 address: 	200:8869:4984:2e39:95b:4eb4:d347:cb0c                           	
IPv6 subnet:  	300:8869:4984:2e39::/64                                         	
Coordinates:  	[]                                                              	
Public key:   	bbcb5b3de8e37b5258a5965c1a79dee07dde372e8cc610e9d23e57d4e8828fd2	
vadym@ubuntu:~/yggdrasil-go$ ./yggdrasilctl -json getnodeinfo key=bbcb5b3de8e37b5258a5965c1a79dee07dde372e8cc610e9d23e57d4e8828fd2
{
  "bbcb5b3de8e37b5258a5965c1a79dee07dde372e8cc610e9d23e57d4e8828fd2": {
    "buildarch": "amd64",
    "buildname": "yggdrasil-develop",
    "buildplatform": "linux",
    "buildversion": "0.4.4-36-g69632ba"
  }
}
```

ygg.conf file:

```
  # Your public key. Your peers may ask you for this to put
  # into their AllowedPublicKeys configuration.
  PublicKey: bbcb5b3de8e37b5258a5965c1a79dee07dde372e8cc610e9d23e57d4e8828fd2

  # Your private key. DO NOT share this with anyone!
  PrivateKey: aa19c2ac9a8f388dda3b4b04365e0d110ce6735c088fcdd1e4bce5d7ba61a83dbbcb5b3de8e37b5258a5965c1a79dee07dde372e8cc610e9d23e57d4e8828fd2

  # Local network interface name for TUN adapter, or "auto" to select
  # an interface automatically, or "none" to run without TUN.
  IfName: auto

  # Maximum Transmission Unit (MTU) size for your local TUN interface.
  # Default is the largest supported size for your platform. The lowest
  # possible value is 1280.
  IfMTU: 65535

  # By default, nodeinfo contains some defaults including the platform,
  # architecture and Yggdrasil version. These can help when surveying
  # the network and diagnosing network routing problems. Enabling
  # nodeinfo privacy prevents this, so that only items specified in
  # "NodeInfo" are sent back if specified.
  NodeInfoPrivacy: false

  # Optional node info. This must be a { "key": "value", ... } map
  # or set as null. This is entirely optional but, if set, is visible
  # to the whole network on request.
  NodeInfo: {
	name: Zurich
  }
}
```

neilalexander:
Still to do before release:

- [x] Fix #954 
- [x] Investigate #891
- [ ] ~~Investigate #953~~


neilalexander:
Release candidate builds: https://nightly.link/yggdrasil-network/yggdrasil-go/actions/runs/3256173796

Vort:
#891 problem is still reproducing as it was before.

neilalexander:
> #891 problem is still reproducing as it was before.

On which version of Windows? I tested it on Server 2019 and can't reproduce it.

Vort:
> On which version of Windows?

On the same as I mentioned in #891 - Windows 7 SP1 x64

neilalexander:
> > On which version of Windows?
> 
> On the same as I mentioned in #891 - Windows 7 SP1 x64

Honestly I don't know what we can do here. I'd need some other people to test it on more recent versions of Windows that are actually in support, as [Windows 7 isn't](https://www.microsoft.com/en-gb/windows/end-of-support).

If we update the Wireguard TUN packages then we will probably break Go 1.17 support and I still don't know if it will fix the problem.

Vort:
I realized that problem is slightly different than I imagined.
(I compare two versions right now: 0.4.0-30-gf92d812 and 0.4.4-41-g6bda57e)
Problem with abrupt exit happens in _both_ versions.
Both versions also leave trash adapter behind.
But in older version it is recreated silently (with 0.11 wintun.dll, driver may be newer).
While in new version (0.14.1 .dll) recreation fails and only new adapter is created (which creates more and more of them with each restart).

Vort:
> Honestly I don't know what we can do here.

Debug prints can be added to code to find where exactly it exits.

> on more recent versions of Windows that are actually in support

It will not fix #891.

> as Windows 7 isn't

OSes do not disappear from existence once support ends.
It is still possible to install them, use and make compatible software for them.

> You can try the attached ZIP which includes the Wireguard dependency update

Same problem happens.

neilalexander:
> Debug prints can be added to code to find where exactly it exits.

There are already `Stopping` and `Stopped` lines in the core router — if those appear in the log, then the TUN adapter has *already been closed*, as `(tun.Device).Close()` from the TUN library has already been called by that point.

> It will not fix #891.

So far I have been unable to reproduce this issue on newer versions of Windows. I don't have a Windows 7 box though, so I would rely on community feedback from other Windows users to understand if the problem is specific to certain versions. Anything is possible in that regard.

> OSes do not disappear from existence once support ends. It is still possible to install them, use and make compatible software for them.

That's true, but there's a limit to what we can reasonably support, especially when we don't own the full supply chain of dependencies. We are only two maintainers doing this entirely in our spare time and we're using libraries written by other people whom we have no control over — if they aren't testing against Windows 7 then the problem probably exists upstream too.

neilalexander:
Updated release candidate builds: https://nightly.link/yggdrasil-network/yggdrasil-go/actions/runs/3256421478

Vort:
> There are already Stopping and Stopped lines in the core router — if those appear in the log

That's the problem - they do not appear in log when I stop service.

neilalexander:
> That's the problem - they do not appear in log when I stop service.

Are you clicking “Stop” or “Restart”? (I don’t know if there is a difference really.)

On Server 2019, they do appear in the log as of today’s builds.

Vort:
> Are you clicking “Stop” or “Restart”?

Stop. From Task Manager by the way. It do not have Restart. But it should not matter.

vikulin:
https://github.com/yggdrasil-network/yggdrasil-go/issues/891 does not reproduce for me: Windows 11, 64bit

PR #958, opened 2022-10-15
Update Wireguard TUN packages

### LINKED ISSUE #891: Incorrect cleanup in service mode

neilalexander:
This might help to resolve #891 maybe?


vikulin:
It doesn't. tested on wintun 0.5.3. The issue is still reproducing on it.

PR #964, opened 2022-10-22
Priority support

neilalexander:
This allows setting priorities for multiple peerings *to the same node*, so that nodes that are multi-homed or have backup connectivity can be specifically configured to prefer certain links over others.

Note that it **does not prioritise peerings to different nodes** and deliberately does not provide the capability to prefer certain upstream nodes over others.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,10 +2,8 @@ module github.com/yggdrasil-network/yggdrasil-go
 
 go 1.17
 
-replace github.com/Arceliar/ironwood => github.com/neilalexander/ironwood v0.0.0-20221019173208-3529562f5a44
-
 require (
-	github.com/Arceliar/ironwood v0.0.0-20220924160422-ed4b6d4750b6
+	github.com/Arceliar/ironwood v0.0.0-20221025225125-45b4281814c2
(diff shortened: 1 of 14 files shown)


Revertron:
> Note that it does not prioritise peerings to different nodes and deliberately does not provide the capability to prefer certain upstream nodes over others.

I hope we will get that in some other PR ;)

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/yggdrasil-network/yggdrasil-go
 
 go 1.17
 
+replace github.com/Arceliar/ironwood => github.com/neilalexander/ironwood v0.0.0-20221019173208-3529562f5a44
```
neilalexander, line 5: Depends on Arceliar/ironwood#5, so needs removing and `go.mod` updating.

neilalexander:
> I hope we will get that in some other PR ;)

That’s a lot less likely at the moment — it’s a lot harder to do that without negatively affecting path stretch or adding in new ways for nodes to potentially cause routing loops.

The only reason this PR is safe is because it doesn’t really affect the routing decisions at all.

PR #977, opened 2022-11-05
make tls certs never expire

### LINKED ISSUE #976: About TLS certificates expiration date
I saw this TODO in the source code:
https://github.com/yggdrasil-network/yggdrasil-go/blob/6112c9cf18929547229e55c9d364efec6f775f88/src/core/link_tls.go#L126-L129
According to the [RFC 5280, Section 4.1.2.5 ("Validity")](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.5), it seems that we can set the expiration date to 9999/12/31 23:59:59 UTC to eliminate the need for rotating the certificates.
```
   In some situations, devices are given certificates for which no good
   expiration date can be assigned.  For example, a device could be
   issued a certificate that binds its model and serial number to its
   public key; such a certificate is intended to be used for the entire
   lifetime of the device.

   To indicate that a certificate has no well-defined expiration date,
   the notAfter SHOULD be assigned the GeneralizedTime value of
   99991231235959Z.
```

majestrate:
according to RFC5280 we can make tls certs never expire by setting their notAfter date to a value that is basically the end of time.

fixes #976

### DIFF
--- a/src/core/link_tls.go
+++ b/src/core/link_tls.go
@@ -120,20 +120,18 @@ func (l *linkTLS) listen(url *url.URL, sintf string) (*Listener, error) {
 	return entry, nil
 }
 
+// RFC5280 section 4.1.2.5
+var notAfterNeverExpires = time.Date(9999, time.December, 31, 23, 59, 59, 0, time.UTC)
+
 func (l *linkTLS) generateConfig() (*tls.Config, error) {
 	certBuf := &bytes.Buffer{}
-
-	// TODO: because NotAfter is finite, we should add some mechanism to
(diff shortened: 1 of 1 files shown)


PufferBlue:
I think it's not the correct date, `UnixMilli(99991231235959)` is actually 5138/08/06 22:00:35.959 UTC
Maybe we can use `time.Date(9999, time.December, 31, 23, 59, 59, 0, time.UTC)` instead

neilalexander (approved):

neilalexander:
Good spot, thanks!

PR #982, opened 2022-11-09
Connect first peer from config on demand.

Revertron:
Now we have a timer that is trying to connect configured peers every minute.
But if you switch network just after that you will need to wait 59 seconds until Yggdrasil will become available again.

I've introduced a new function that will reconnect the first (and often only) peer from config on demand. It will be called from Kotlin side when network becomes available.

(Patch for yggdrasil-android will follow)


neilalexander:
No longer needed after #983.

PR #1029, opened 2023-04-17
Add output for threadcount and key generation time to cmd/genkey

jjolly:
This change is to display information about the key generation process.

Specifically, two bits of information are now displayed
 * The number of threads created to search for keys, and
 * The time taken to generate a successful "next best" key


jjolly:
This is just a simple vanity change. I wanted to see how key generation performed between machines. It fixes no bugs and adds no new functionality, but I though I would share something useful to me.

Issue #1031, opened 2023-04-24
[Admin API] Router response does not contain request body


one-d-wide:
Router response does not contain request body, as it states [documentation](https://github.com/yggdrasil-network/yggdrasil-network.github.io/blob/0cfad2ee414257f690cbcf4f9287261d9b07e3c1/admin.md?plain=1#L84). This bug prohibits concurrent queries, because they can't be distinguished.

```
$ nc -U /var/run/yggdrasil/yggdrasil.sock
{
  "request": "getself",
  "arguments": {},
  "my_nonce": 1234
} 
{
  "status": "success",
  "request": null, <-- should be the request body
  "response": {
    "build_name": "yggdrasil",
    "build_version": "0.4.7",
    "key": "<my_key>",
    "address": "<my_addr>",
    "coords": [
      <my_cords>,
      <...>
    ],
    "subnet": "<my_subnet>"
  })
}
```

UPDATE:
I incidentally noticed that my older router (v0.4.4) actually sets `request` field correctly, although without custom `my_nonce`. After a bit of research I found that earlier request value [was parsed into](https://github.com/yggdrasil-network/yggdrasil-go/blob/c6fe81b5d282ae8166b373e1920b26676dcf5aac/src/admin/admin.go#L285) the `Request` field of variable `resp: AdminSocketResponse` for further processing, but after [refatoring](https://github.com/yggdrasil-network/yggdrasil-go/commit/c922eba2d82503e6ef187d3d3f06572d4b0bacec#diff-74c0673b7d59cfa5721de9c727ede8ae1b87b5223e8b1d796a391105681fed39L306) local variable `res` was used instead, without filling `resp.Request`.

neilalexander:
Fixed in v0.5.

PR #1032, opened 2023-04-28
[Admin API] Fill `Request` attribute in router response

one-d-wide:
Fix for #1031


neilalexander:
Thanks for this, but I'm going to close it as it's already fixed on the `future` branch in commit 7afa23b which will get merged in probably soon.

PR #1036, opened 2023-05-20
Tweak link state locking, add comments, listener priority, other fixes

neilalexander:
This uses a mutex instead of an actor for the link state as it is easier to reason about for now and changes a few other things.

### DIFF
--- a/src/core/link.go
+++ b/src/core/link.go
@@ -121,16 +121,6 @@ func (l *links) shutdown() {
 	})
 }
 
-func (l *links) isConnectedTo(info linkInfo) bool {
-	l.RLock()
-	link, ok := l._links[info]
-	l.RUnlock()
-	if !ok {
-		return false
-	}
-	return link._conn != nil
-}
-
 type linkError string
 
 func (e linkError) Error() string { return string(e) }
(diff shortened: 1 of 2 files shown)


src/core/link.go
```diff
@@ -149,10 +145,14 @@ func (l *links) add(u *url.URL, sintf string, linkType linkType) error {
...
+		uri:   lu.String(),
+		sintf: sintf,
 	}
+
+	// If we think we're already connected to this peer, load up
+	// the existing peer state. Try to kick the peer if possible,
+	// which will cause an immediate connection attempt if it is
+	// backing off for some reason.
 	l.RLock()
 	state, ok := l._links[info]
```
Arceliar, line 157: I think this races with the write later in this function.

src/core/link.go
```diff
@@ -198,8 +196,7 @@ func (l *links) add(u *url.URL, sintf string, linkType linkType) error {
 		options.priority = uint8(pi)
 	}
 
-	// Store the state of the link, try to connect and then run
-	// the handler.
+	// Store the state of the link so that it can be queried later.
 	l.Lock()
 	l._links[info] = state
```
Arceliar, line 201: I think this races with the above read.

src/core/link.go
```diff
@@ -324,41 +362,68 @@ func (l *links) listen(u *url.URL, sintf string) (*Listener, error) {
...
+				// If this node is already connected to us, just drop the
+				// connection. This prevents duplicate peerings.
 				if l.isConnectedTo(info) {
 					return
 				}
+
+				// If there's an existing link state for this link, get it.
+				// Otherwise just create a new one.
 				l.RLock()
 				state, ok := l._links[info]
```
Arceliar, line 386: I think this races with the write below.

src/core/link.go
```diff
@@ -324,41 +362,68 @@ func (l *links) listen(u *url.URL, sintf string) (*Listener, error) {
...
+				// Clear the error state.
+				state.Lock()
+				state._conn = lc
+				state._err = nil
+				state._errtime = time.Time{}
+				state.Unlock()
+
+				// Store the state of the link so that it can be queried later.
 				l.Lock()
 				l._links[info] = state
```
Arceliar, line 414: I think this races with the above read.

Arceliar (changes requested):
I think there are a few races in how these locks are used. Multiple calls to `add` or `listen` can conflict with each other, so they overwrite state that may have been added after the initial read/check was done. I would suggest grabbing the write lock at that initial read and holding it until either the write is done or we see that there's something already present.

I'm not 100% sure the semantics of the rest of the code base would ever trigger this (maybe if we have incoming/outgoing connections to the same multicast peer happening at the same time), but it's probably best to be conservative and hold locks long enough to guarantee we're safe no matter how things are called.

Arceliar (approved):
LGTM

PR #1043, opened 2023-06-08
core/link_tcp_linux: drop trying to set BBR

pfactum:
Instead of implementing an ability to select arbitrary TCP congestion control algorithm (#776), drop setting it altogether as suggested by @link2xt in #754.

Link: https://github.com/yggdrasil-network/yggdrasil-go/pull/776
Fixes: https://github.com/yggdrasil-network/yggdrasil-go/issues/754


pfactum:
Eh, given https://github.com/yggdrasil-network/yggdrasil-go/commit/fa3d943ba90f168d3cc591c276b944f1eeda1344, I guess this should be closed now. Although I'm not sure why this PR was ignored.

PR #1045, opened 2023-06-18
Define interface for RWCs

neilalexander:
This means that we no longer need to fork `tun` to swap out RWCs.


Arceliar (approved):
LGTM (assuming CI passes).

PR #1046, opened 2023-06-18
Tweak link handshake

neilalexander:
The handshake format is already mostly extensible but including the total length of the data in the handshake lets us read the exact amount of bytes from the wire precisely. Previously both the sent and received handshakes had to be the same length.


Arceliar (approved):
LGTM

PR #1052, opened 2023-08-12
Multipath TCP

neilalexander:
This adds support for using MPTCP with Go 1.21 on platforms that support it.


cathugger:
I'd probably gate auto-enable on kernel version.
According to https://github.com/multipath-tcp/mptcp_net-next/wiki#changelog there's been a lot of changes and initial versions had a lot missing.
Something like v5.19 or maybe v6.1 IMO makes sense.

AlexFreeman13:
Please upgrade mptcp branch to 0.5.7

PR #1073, opened 2023-10-27
Debian package updates

neilalexander:
This makes the following changes to the Debian package:
* The config is now in `/etc/yggdrasil/yggdrasil.conf`
* The admin socket is now in `/var/run/yggdrasil/yggdrasil.sock`
* Generated configuration files will no longer contain the `AdminListen` option by default, as that way the platform default can be used instead
* Existing configurations in `/etc/yggdrasil.conf` will be moved to `/etc/yggdrasil/yggdrasil.conf`
* A number of additional protections are now present in the systemd unit file


contrib/systemd/yggdrasil.service
```diff
@@ -8,14 +8,18 @@ After=yggdrasil-default-config.service
...
+ProtectSystem=strict
+NoNewPrivileges=true
+RuntimeDirectory=yggdrasil
+ReadWritePaths=/var/run/yggdrasil/ /run/yggdrasil/
 SyslogIdentifier=yggdrasil
 CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_BIND_SERVICE
+AmbientCapabilities=CAP_NET_ADMIN CAP_NET_BIND_SERVICE
 ExecStartPre=+-/sbin/modprobe tun
-ExecStart=/usr/bin/yggdrasil -useconffile /etc/yggdrasil.conf
+ExecStart=/usr/bin/yggdrasil -useconffile /etc/yggdrasil.conf.backup
```
Revertron, line 19: `.backup`? Why? :)

neilalexander: Good catch, that was from testing.

PR #1088, opened 2023-11-03
Re-add `PublicKey` to configuration file

neilalexander:
This PR does two things:
* Re-adds the `PublicKey` to the configuration file — technically it does nothing but `-normaliseconf` will update it
* Cuts down `PrivateKey` to the seed length only and updates things to expect the seed only


neilalexander:
We've decided against having effectively a no-op configuration line. I will add a `-publickey` command line option in a separate PR.

PR #1089, opened 2023-11-03
Don't panic when connect returns `nil` (fixes #1086)

### LINKED ISSUE #1086: Arch Linux update from 0.4.7 to 0.5.1: panic: runtime error: invalid memory address or nil pointer dereference
Hello,
the Arch Linux package was finally updated to 0.5.1 today, so I went and updated my public node and personal node, and in both cases I receive the same issue:

upon launching yggdrasil with the new version using the old config file, I get the following error:

```bash
$ sudo yggdrasil -useconffile /etc/yggdrasil.conf
2023/11/03 12:47:16 Build name: yggdrasil
2023/11/03 12:47:16 Build version: 0.5.1
2023/11/03 12:47:16 Failed to connect to [2001:19f0:5:2c95:5400:1ff:fe46:9c]:34660: dial tcp [2001:19f0:5:2c95:5400:1ff:fe46:9c]:34660: connect: network is unreachable
2023/11/03 12:47:16 Failed to connect to [2001:19f0:5:6ed:5400:ff:fe5f:94f7]:12345: dial tcp [2001:19f0:5:6ed:5400:ff:fe5f:94f7]:12345: connect: network is unreachable
2023/11/03 12:47:16 Failed to connect to [2001:19f0:5:2a83:5400:1ff:fe89:44db]:58149: dial tcp [2001:19f0:5:2a83:5400:1ff:fe89:44db]:58149: connect: network is unreachable
2023/11/03 12:47:16 Failed to connect to [2604:a880:cad:d0::45:d001]:59168: dial tcp [2604:a880:cad:d0::45:d001]:59168: connect: network is unreachable
2023/11/03 12:47:16 Failed to connect to [2604:180:1:568::beef]:12345: dial tcp [2604:180:1:568::beef]:12345: connect: network is unreachable
2023/11/03 12:47:16 Failed to connect to [2600:3c00::f03c:91ff:feae:3efa]:44478: dial tcp [2600:3c00::f03c:91ff:feae:3efa]:44478: connect: network is unreachable
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x55b7f026cee4]

goroutine 56 [running]:
github.com/yggdrasil-network/yggdrasil-go/src/core.(*linkConn).SetDeadline(0xc14966e696d420e2?, {0x165ce249e?, 0x55b7f06c4b60?, 0x55b7f06c4b60?})
    <autogenerated>:1 +0x24
github.com/yggdrasil-network/yggdrasil-go/src/core.(*links).handler(0xc0000021e8, 0x0, {0x0, 0x0, {0x0, 0x0}, {0x0, 0x0, 0x0}}, {0x55b7f042e5e8, ...})
    github.com/yggdrasil-network/yggdrasil-go/src/core/link.go:537 +0x18c
github.com/yggdrasil-network/yggdrasil-go/src/core.(*links).add.func1.2()
    github.com/yggdrasil-network/yggdrasil-go/src/core/link.go:309 +0x4f9
created by github.com/yggdrasil-network/yggdrasil-go/src/core.(*links).add.func1 in goroutine 55
    github.com/yggdrasil-network/yggdrasil-go/src/core/link.go:240 +0x809
```

I've tried using the -normaliseconf option from the configuration documentation to update it (by replacing the old one with the normalized one), and the error doesn't leave, however, everything seems to be fine if I start it with autoconf, which makes it useless as you can't have any peers set up from the start in that case. I'm not sure what else to try to fix it. Any help would be appreciated.

Thank you.

neilalexander:
It isn't clear to me why this would happen but let's guard the condition anyway.

### DIFF
--- a/src/core/link.go
+++ b/src/core/link.go
@@ -274,6 +274,9 @@ func (l *links) add(u *url.URL, sintf string, linkType linkType) error {
 
 				conn, err := l.connect(state.ctx, u, info, options)
 				if err != nil || conn == nil {
+					if err == nil && conn == nil {
+						l.core.log.Warnf("Link %q reached inconsistent error state", u.String())
+					}
 					if linkType == linkTypePersistent {
 						// If the link is a persistent configured peering,
 						// store information about the connection error so
--- a/src/core/link.go
+++ b/src/core/link.go
@@ -273,7 +273,7 @@ func (l *links) add(u *url.URL, sintf string, linkType linkType) error {
 				}
 
 				conn, err := l.connect(state.ctx, u, info, options)
-				if err != nil {
+				if err != nil || conn == nil {
 					if linkType == linkTypePersistent {
 						// If the link is a persistent configured peering,
 						// store information about the connection error so


neilalexander:
@Kusoneko If you can try this branch and let me know if it helps, I'd appreciate it.

Kusoneko:
I can confirm that it appears to work using this branch (at least for the few minutes I ran it manually, which is more than I can say of 0.5.1 which just crashes in less than a second), I'm guessing this means the issue is I have some peers in both of my machines' yggdrasil.conf that are either offline or something along those lines and it makes it crash?

neilalexander:
Peers being offline _shouldn't_ make it crash, at least I haven't seen that on my own nodes, but clearly something is up. Thanks for testing it though, this will go into v0.5.2.

src/core/link.go
```diff
@@ -257,7 +257,7 @@ func (l *links) add(u *url.URL, sintf string, linkType linkType) error {
 				}
 
 				conn, err := l.connect(state.ctx, u, info, options)
-				if err != nil {
+				if err != nil || conn == nil {
```
Arceliar, line 260: Since a nil error and a nil conn is weird, we should probably check if conn == nil and add some additional logging, then ask someone who sees the error to run it. E.g. at least see which protocol the peers are trying to connect over (maybe it's some edge case in the QUIC code?).

Arceliar (changes requested):
We should add some logging for the `conn == nil` case to try to figure out what the underlying problem is.

src/core/link.go
neilalexander: Added this.

Arceliar (approved):

Kusoneko:
I went and ran it for a bit for the logging, I'm fairly certain I haven't updated my peers list for either of my machines for a few years now, I probably should, but the "having to do it manually" part is a bit painful. Anyways, here's my local yggdrasil.conf (redacted to remove private key info and the comments) and the output of a few minutes of running it:

[yggdrasil.conf](https://github.com/yggdrasil-network/yggdrasil-go/files/13257605/yggdrasil.conf.txt)
[log.txt](https://github.com/yggdrasil-network/yggdrasil-go/files/13257606/log.txt)

Here are the ones for my public peer:

[yggdrasil.conf.txt](https://github.com/yggdrasil-network/yggdrasil-go/files/13257629/yggdrasil.conf.txt)
[log.txt](https://github.com/yggdrasil-network/yggdrasil-go/files/13257630/log.txt)

Hopefully all this helps figure out why this bug even happened to begin with.

As an aside, a way to quickly update the peers/interfacepeers part of the conf file to include all public peers would be nice, as well as an option to update it to the, say, 2-4 fastest peers -- and remove peers that no longer work. (As far as I can tell there isn't any way but manually at the moment?) I'm aware the recommendation leans more towards the latter, which is what I should do for my private machine, but for my public peer, I don't really mind it connecting to everything and having traffic route through it. Currently having to update those manually leads to me preferring the riskier approach of adding more peers to each of my machines to avoid having to update the config file manually for a long time as opposed to regularly updating it every time a new commit lands in the public peers repo, which I usually don't look at religiously and forget it exists for long periods of time, at least until someone mentions me there to tell me my public peer broke or something.

PR #1090, opened 2023-11-03
Add `sockstls://`

### LINKED ISSUE #1087: Feature request: socks+tls
TLS is the most secure option to connect the nodes, now it got even better with that password feature which probably makes it impossible to MITM yggdrasil connection to the node (i know its encrypted end2end but attacker on ISP side can at least fuck up connection and routes like that).

But I'm really missing socks option for TLS. For some reason yggdrasil socks option can only work with TCP, and sometimes its pretty convenient to add bunch of other gateways this way.

Nothing urgent here but i just wonder why socks+tls doesnt exists, its still TCP and can be wrapped into socks same way. QUIC on the other hand would be tricky to route over socks (with all those UDP bind tricks), but TLS supposed to be pretty trivial.

Btw: If you implement socks+QUIC via UDP bind it should be even faster, because UDP bind is just userspace port forwarding and causes zero processing lags compared to TCP via socks which endures encapsulation and routing overheads inside socks server. Openvpn was able to do udpbind via socks and it was pretty much fast, not distinguishable from direct UDP connection.

neilalexander:
Closes #1087. Allows connecting to a `tls://` peer over SOCKS, whereas the standard `socks://` scheme can only connect to `tcp://` peers.

### DIFF
--- a/src/core/link.go
+++ b/src/core/link.go
@@ -179,6 +179,22 @@ func (l *links) add(u *url.URL, sintf string, linkType linkType) error {
 			}
 			options.password = []byte(p)
 		}
+		// SNI headers must contain hostnames and not IP addresses, so we must make sure
+		// that we do not populate the SNI with an IP literal. We do this by splitting
+		// the host-port combo from the query option and then seeing if it parses to an
+		// IP address successfully or not.
+		if sni := u.Query().Get("sni"); sni != "" {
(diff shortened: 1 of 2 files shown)


Arceliar (approved):
Not tested to see if it works when enabled, but it doesn't seem to break anything when not used, so we can iterate on this if need be.

PR #1109, opened 2023-11-26
Port netstack branch to latest develop

basilgello:
yggstack is really useful in some situations, namely in Termux.
Old branch does not work with 0.5.xso I ported it to support new protocol.

Tested on:

* Yggstack client (socks) <-> Yggdrasil server (tun)
* Yggstack client (socks) <-> Yggstack server (exposetcp)

Also I had to switch back to upstream netstack and do couple of minor logging fixes.


majestrate:
neato. will review this once it goes out of draft phase.

src/types/tcpproxy.go
```diff
@@ -0,0 +1,43 @@
+package types
+
+import "net"
+
+func connToChan(mtu uint64, conn net.Conn) chan []byte {
+	c := make(chan []byte)
+	go func() {
+		for {
+			b := make([]byte, mtu)
```
majestrate, line 9: would this chug gc less if we used a global `sync.Pool` here?

majestrate: actually, what if we used `io.Pipe()` and `io.CopyN()` in `ProxyTCP()`?

majestrate: actually, you don't even need `io.Pipe()` you could just do `io.CopyBuffer` and flip which connection is passed in. no need to use a channel. unless i am missing something.

cmd/yggdrasil/main.go
```diff
@@ -2,57 +2,26 @@ package main
...
 
-	"github.com/yggdrasil-network/yggdrasil-go/src/address"
-	"github.com/yggdrasil-network/yggdrasil-go/src/admin"
 	"github.com/yggdrasil-network/yggdrasil-go/src/config"
-	"github.com/yggdrasil-network/yggdrasil-go/src/ipv6rwc"
-
-	"github.com/yggdrasil-network/yggdrasil-go/src/core"
-	"github.com/yggdrasil-network/yggdrasil-go/src/multicast"
-	"github.com/yggdrasil-network/yggdrasil-go/src/tun"
+	"github.com/yggdrasil-network/yggdrasil-go/src/setup"
```
majestrate, line 18: anecdote: this split is something that i've wanted for a while for my own reasons for a while.

skyguy (commented):
Can the resolver split off the prefix to enable virtual hosts?

cmd/yggdrasil/main.go
basilgello: @neilalexander You told me on Matrix that setup split is not needed. @majestrate what's your reason?

cmd/yggstack/types/resolver.go
```diff
@@ -31,6 +37,16 @@ func NewNameResolver(stack *netstack.YggdrasilNetstack, nameserver string) *Name
 }
 
 func (r *NameResolver) Resolve(ctx context.Context, name string) (context.Context, net.IP, error) {
+	if strings.HasSuffix(name, NameMappingSuffix) {
+		name = strings.TrimSuffix(name, NameMappingSuffix)
```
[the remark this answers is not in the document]
basilgello: Doable

cmd/yggdrasil/main.go
majestrate: it seems to make initializing the core much simpler. i also attempted to do a netstack version of ygg a while back and that lack of split was a burden for me.

basilgello:
Addressed all the reviee points till now and pushed to https://github.com/yggdrasil-network/yggstack

PR #1116, opened 2023-12-20
Update rc script for FreeBSD and fix logging

mh-56:
This commit updates rc script for FreeBSD and fixes logging. syslog is now used and messages are saved to /var/log/yggdrasil.log instead of empty /var/log/yggdrasil.stdout.log and /var/log/yggdrasil.stderr.log. Log rotation has been added via newsyslog.

### DIFF
--- a/contrib/freebsd/yggdrasil
+++ b/contrib/freebsd/yggdrasil
@@ -1,10 +1,12 @@
 #!/bin/sh
 #
 # Put the yggdrasil and yggdrasilctl binaries into /usr/local/bin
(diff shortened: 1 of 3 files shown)


neilalexander:
If it is the FreeBSD package in particular that you want to fix, take a look at https://github.com/yggdrasil-network/yggdrasil-package-freebsd as it is more up-to-date than what's in `contrib`.

mh-56:
Committed in yggdrasil-package-freebsd

PR #1124, opened 2024-01-15
Adjust default backoff max to just over 1 hour, add `?maxbackoff=` peer option

### LINKED ISSUE #1117: Allow a configurable maximum backoff
Hello,

First, thanks for all the work on Yggdrasil!

In Yggdrasil 0.5, an exponential backoff was added, probably related to #968 .  This makes sense in many contexts.

However, the documented maximum backoff (4+ hours) is so high that it causes problems.  In some cases, I run my own peers, and seconds matter.  I would like them to reconnect ASAP.

In other cases, I have a system with intermittent LTE connectivity.  It may be down for an hour or two due to lack of a cell signal, but the moment it can get through again (or soon thereafter), I want it to connect.  If it waits 20 minutes, or 4 hours, it may well have missed its opportunity.

So, one or more of these three things might be useful:

1. A configurable maximum backoff limit per-peer, using a mechanism similar to the `?priority=` we already have
2. A global maximum backoff
3. A lower hardcoded maximum backoff

Thanks!

neilalexander:
This makes two changes to the link backoffs:

* The default maximum is now just over 1 hour, instead of 4.5 hours.
* A new peer option `?maxbackoff=` is added, minimum allowed value is `30s`, accepts minutes like `5m` or hours like `4h` etc. The backoff is exponential as normal up until that limit is reached, at which point the backoff is held at that limit.

Fixes #1117.

### DIFF
--- a/src/core/link.go
+++ b/src/core/link.go
@@ -6,7 +6,6 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
-	"math"
 	"net"
 	"net/netip"
 	"net/url"
@@ -27,6 +26,9 @@ const (
 	linkTypeEphemeral                  // Multicast discovered
 	linkTypeIncoming                   // Incoming connection
 )
+
+const defaultBackoffLimit = time.Second << 12 // 1h8m16s
+const minimumBackoffLimit = time.Second * 30
 
 type links struct {
 	phony.Inbox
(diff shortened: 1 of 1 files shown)


Arceliar (approved):

lostmsu:
Why is there a minimum allowed value? As in https://github.com/yggdrasil-network/yggdrasil-go/issues/1117 I have my own peers, and I am building a distributed system with many components, so each component taking its time can significantly increase overall recovery time of the system.

PR #1131, opened 2024-02-26
Replacing wix implementation for MSI builds

Saiv46:
A little experiment using `wixl` from Ubuntu packages. Built `.msi` works, but ARM is not supported.


neilalexander:
For what it's worth, we've been here before (#766) and then reverted it due to problems with the package and Wintun.

PR #1145, opened 2024-05-31
TUN vectorised reads/writes

neilalexander:
This PR updates the Wireguard dependency and updates to use new vectorised reads/writes, which should reduce the number of syscalls and improve performance.

This will only make a difference on Linux as this is the only platform for which the Wireguard TUN library supports vectorised reads/writes. For other platforms, single reads and writes will be performed as usual.

### DIFF
--- a/src/tun/iface.go
+++ b/src/tun/iface.go
@@ -1,33 +1,48 @@
 package tun
 
-const TUN_OFFSET_BYTES = 4
+import (
+	"net"
+)
+
+const TUN_OFFSET_BYTES = 80 // sizeof(virtio_net_hdr)
+const TUN_MAX_VECTOR = 16
+
+func (tun *TunAdapter) idealBatchSize() int {
+	if b := tun.iface.BatchSize(); b <= TUN_MAX_VECTOR {
+		return b
+	}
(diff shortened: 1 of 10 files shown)


src/tun/iface.go
```diff
@@ -1,42 +1,64 @@
 package tun
 
-const TUN_OFFSET_BYTES = 4
+const TUN_OFFSET_BYTES = 80 // sizeof(virtio_net_hdr)
+
+func (tun *TunAdapter) idealBatchSize() int {
+	return tun.iface.BatchSize()
```
Arceliar, line 6: Minor point, but now that this is just calling `tun.iface.BatchSize()`, we could pretty easily remove this function and call that directly. Unless you think `idealBatchSize` helps with readability, I'm pretty in different to this, so I'm fine with leaving it in that case.

Arceliar (approved):
LGTM modulo one function that the last commit turned into a simple method chain, so we could think about removing that and calling the underlying function directly. I'm fine with either approach, I just don't want it to be a historical accident if we keep it.

src/tun/iface.go
neilalexander: Fixed.

PR #1148, opened 2024-06-15
Notify service manager when start up is completed

one-d-wide:
Currently there is no way to notify service manager when yggdrasil start up process is truly completed, i.e its TUN interface and admin API endpoints are up and running. This makes naive services, set to depend on the yggdrasil network, periodically crash on startup, requiring sub-optimal complications for trivial tasks.

This PR solves the issue by notifying the service manager after TUN interface and admin sockets are ready. Systemd support is implemented by sending a short `READY=1` message to a UNIX datagram socket defined in a `NOTIFY_SOCKET` environment variable, see `sd_notify(3)`.

### Changes

- `cmd/yggdrasil/main.go` - Added invocation for `notifyStartupCompleted` right after TUN interface and admin sockets are set up.
- `cmd/yggdrasil/notify_startup_*.go` - Added `notifyStartupCompleted` function handling the process on linux and immediately returning on other platforms.
- `contrib/systemd/yggdrasil.service*` - Service `Type` is changed to `notify`, added a comment about `NotifyAccess` property.

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -272,6 +272,10 @@ func main() {
 		}
 	}
 
+	if _, err = notifyStartupCompleted(); err != nil {
+		log.Warnln("Error while sending start up notification:", err)
+	}
+
 	// Block until we are told to shut down.
 	<-ctx.Done()
 
--- a/cmd/yggdrasil/notify_startup_linux.go
+++ b/cmd/yggdrasil/notify_startup_linux.go
@@ -1,0 +1,50 @@
+//go:build linux
+// +build linux
+
+package main
+
+import (
+	"net"
+	"os"
+)
+
+// Notify systemd daemon when start up is completed.
+// Required to ensure that dependent services are started only after TUN interface is ready.
+//
+// One of the following is returned:
+// (false, nil) - notification not supported (i.e. `notifySocketEnv` is unset)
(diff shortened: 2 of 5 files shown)


neilalexander:
Not intentionally ignoring this PR BTW, but it has a knock-on effect to quite a few packages, so probably not for a patch release.

cmd/yggdrasil/main.go
```diff
@@ -272,6 +272,10 @@ func main() {
 		}
 	}
 
+	if _, err = notifyStartupCompleted(); err != nil {
+		log.Warnln("Error while sending start up notification:", err)
+	}
+
```
klemensn, line 278: There is no guarantee at this point (in the main goroutine) that startup tasks in other goroutines have been finished, since they don't synchronise.

So I think you are just racing against tun(4) setup just like `-chuser` is already racing against the control socket setup over at https://github.com/yggdrasil-network/yggdrasil-go/pull/927#discussion_r1826606055

cmd/yggdrasil/main.go
```diff
@@ -272,6 +272,10 @@ func main() {
 		}
 	}
 
+	if _, err = notifyStartupCompleted(); err != nil {
```
klemensn, line 275: The first `bool` return value is never used, so why have it at all?

PR #1152, opened 2024-07-20
Implement websocket (ws:// and wss://) links

basilgello:
ws:// can be listened and dialed
wss:// is a convenience link for ws:// that supports dialing to ws:// peer.

### DIFF
--- a/src/core/link_ws.go
+++ b/src/core/link_ws.go
@@ -48,20 +48,19 @@ func (l *linkWSListener) Close() error {
 	if err := l.httpServer.Shutdown(l.ctx); err != nil {
 		return err
 	}
-
 	return l.listener.Close()
 }
 
 func (s *wsServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.URL.Path == "/h" {
+	if r.URL.Path == "/health" {
 		w.WriteHeader(http.StatusOK)
-		w.Write([]byte("OK"))
+		_, _ = w.Write([]byte("OK"))
 		return
 	}
+
 	c, err := websocket.Accept(w, r, &websocket.AcceptOptions{
 		Subprotocols: []string{"ygg-ws"},
 	})
-
 	if err != nil {
 		return
 	}
@@ -71,11 +70,8 @@ func (s *wsServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	netconn := websocket.NetConn(s.ctx, c, websocket.MessageBinary)
-
-	ch := s.ch
-	ch <- &linkWSConn{
-		Conn: netconn,
+	s.ch <- &linkWSConn{
+		Conn: websocket.NetConn(s.ctx, c, websocket.MessageBinary),
 	}
 }
 
@@ -83,7 +79,6 @@ func (l *links) newLinkWS() *linkWS {
 	lt := &linkWS{
 		links: l,
 	}
-
 	return lt
 }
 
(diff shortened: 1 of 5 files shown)


src/core/link_ws.go
```diff
@@ -0,0 +1,216 @@
...
+type linkWSS struct {
+	*linkWS
+	tlsconfig  *tls.Config
+}
+
+type linkWSConn struct {
+	net.Conn
+}
+
+type linkWSListener struct {
```
neilalexander, line 29: Looks like this file isn't `gofmt`'d?

src/core/link_ws.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (l *linkWSListener) Addr() net.Addr {
+	return l.listener.Addr()
+}
+
+func (l *linkWSSListener) Addr() net.Addr {
+	return l.listener.Addr()
+}
+
+func (l *linkWSListener) Close() error {
```
neilalexander, line 73: Can shorten to something like:
```
{
  if err := l.httpServer.Shutdown(l.ctx); err != nil {
    return err
  }
  return l.listener.Close()
}
```

src/core/link_ws.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = l.listener.Close()
+
+	return err
+}
+
+func (l *linkWSSListener) Close() error {
```
neilalexander, line 87: Likewise here:
```
{
  if err := l.httpServer.Shutdown(l.ctx); err != nil {
    return err
  }
  if err := l.tlslistener.Close(); err != nil {
    return err
  }
  return l.listener.Close()
}
```

src/core/link_ws.go
```diff
@@ -0,0 +1,216 @@
...
+	return err
+}
+
+func (s *wsServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	c, err := websocket.Accept(w, r, &websocket.AcceptOptions{
+		Subprotocols: []string{"ygg-ws"},
+	})
+
+	if err != nil {
+		return
```
neilalexander, line 113: Do we want to return an error code in this situation? i.e.
```
  if err != nil {
    w.WriteHeader(http.StatusUpgradeRequired)
    return
  }
```
... or is the WebSocket library doing that automatically for us?

neilalexander (changes requested):
Principally fine, might be best to split into `link_ws.go` and `link_wss.go` for brevity/clarity, but a few inline comments below.

src/core/link_ws.go
basilgello: Yup, my fault :)

src/core/link_ws.go
basilgello: Done

src/core/link_ws.go
basilgello: Done

src/core/link_ws.go
basilgello: Yes, it does one for us and this is mentioned in author's "production quality demo project" I used as source

src/core/link_ws.go
```diff
@@ -0,0 +1,216 @@
...
+	lwss := &linkWSS{
+		linkWS:    lws,
+		tlsconfig: l.core.config.tls.Clone(),
+	}
+
+	return lwss
+}
+
+func (l *linkWS) dial(ctx context.Context, url *url.URL, info linkInfo, options linkOptions) (net.Conn, error) {
+	wsconn, _, err := websocket.Dial(ctx, url.Scheme + "://" + url.Host, &websocket.DialOptions{
```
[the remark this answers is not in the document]
basilgello: Makes sense

basilgello:
LGTM :)

Issue #1098, opened 2023-11-06
Silent crashes in Windows service mode


Vort:
Today my node crashed.
Only information about this event was here:
![image](https://github.com/yggdrasil-network/yggdrasil-go/assets/1242858/9deb5a83-a177-4dbd-b59c-4fbed54619b4)
Log file contained absolutely nothing which may help to figure out what problem is.
Please provide way to get additional information about crash when Yggdrasil is running in Windows service mode.

This issue is not about particular crash, it may be fixed with latest commits and I updated my node to `0.5.1-0010` already.

devinterx:
same

Revertron:
Fixed.

Vort:
@Revertron what will happen with your fix in case of crash? Will information about it be logged to the log file?

Revertron:
I've fixed the crash, not the logging.

Vort:
Ok.

> This issue is not about particular crash

neilalexander:
I am not sure this is fixable BTW. When a Go process panics, the error logging is written to `stderr` and there's nothing we can do to control that behaviour. If the Windows service manager doesn't capture that data anywhere then it's unavoidably lost.

Vort:
Quick googling suggests that `stderr` can be redirected with `freopen` or `SetStdHandle` calls.

Issue #993, opened 2022-11-22
Windows version fails to start


Revertron:
I've found time to debug this one :)

When started from admin command line it shows this error:
```
2022/11/23 00:27:09 Build name: yggdrasil-develop
2022/11/23 00:27:09 Build version: 0.4.7
2022/11/23 00:27:09 TCP admin socket listening on 127.0.0.1:9001
2022/11/23 00:27:09 Connected outbound TCP: blablabla
2022/11/23 00:27:09 Using existing driver 0.14
2022/11/23 00:27:09 Creating adapter
2022/11/23 00:27:09 Connected outbound TLS: blablabla
2022/11/23 00:27:24 Timed out waiting for device query: The wait operation timed out. (Code 0x00000102)
2022/11/23 00:27:24 Failed to setup adapter (problem code: 0x1F, ntstatus: 0xC0000035): Cannot create a file when that file already exists. (Code 0x000000B7)
panic: Error creating interface: Cannot create a file when that file already exists.

goroutine 21 [running]:
main.run({0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, {0xc0000bc0c0, 0x27}, ...}, ...)
        /home/revertron/IdeaProjects/yggdrasil-go/cmd/yggdrasil/main.go:365 +0x1585
main.main.func1()
        /home/revertron/IdeaProjects/yggdrasil-go/cmd/yggdrasil/main.go:405 +0x85
created by main.main
        /home/revertron/IdeaProjects/yggdrasil-go/cmd/yggdrasil/main.go:403 +0x1b7
```
But sometimes it just starts instantaneously:
```
2022/11/23 00:31:08 Build name: yggdrasil-develop
2022/11/23 00:31:08 Build version: 0.4.7
2022/11/23 00:31:08 TCP admin socket listening on 127.0.0.1:9001
2022/11/23 00:31:08 Connected outbound TCP: blablabla
2022/11/23 00:31:08 Using existing driver 0.14
2022/11/23 00:31:08 Creating adapter
2022/11/23 00:31:08 Your public key is blablabla
2022/11/23 00:31:08 Your IPv6 address is 200:blablabla
2022/11/23 00:31:08 Your IPv6 subnet is 300:blablabla
2022/11/23 00:31:09 Connected outbound TLS: blablabla
```

POMATu:
the issue can be solved with windows service to be set as 
- Restart time - 1 second
[...]

PR #1153, opened 2024-07-24
Fixed Windows service life-cycle.

Revertron:
This fix fixes two issues: https://github.com/yggdrasil-network/yggdrasil-go/issues/993 & https://github.com/yggdrasil-network/yggdrasil-go/issues/1098

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"crypto/ed25519"
 	"encoding/hex"
 	"encoding/json"
@@ -55,8 +56,9 @@ func main() {
 	
 	done := make(chan struct{})
 	defer close(done)
-	sigCh := make(chan os.Signal, 1)
-	signal.Notify(sigCh, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+	
+	// Catch interrupts from the operating system to exit gracefully.
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 
 	// Create a new logger that logs output to stdout.
 	var logger *log.Logger
@@ -273,13 +275,13 @@ func main() {
 	//Windows service shutdown
 	minwinsvc.SetOnExit(func() {
 		logger.Infof("Shutting down service ...")
-		sigCh <- os.Interrupt
+		cancel()
 		// Wait for all parts to shutdown properly
 		<-done
 	})
 
 	// Block until we are told to shut down.
-	<-sigCh
+	<-ctx.Done()
 
 	// Shut down the node.
 	_ = n.admin.Stop()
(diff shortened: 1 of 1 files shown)


cmd/yggdrasil/main.go
```diff
@@ -271,15 +272,53 @@ func main() {
...
+		// Wait for all parts to shutdown properly
+		<-doneCh
+	})
 
 	// Block until we are told to shut down.
-	<-ctx.Done()
+	<-sigCh
+
+	// Shut down the node using a wait group to synchronize
+	var wg sync.WaitGroup
```
neilalexander, line 288: Bit confused about what all this achieves, as all three `Stop()` functions block until that module is shut down anyway and therefore run sequentially.

This also just means that potentially, depending on scheduling, that `core.Stop()` can run before the others do which doesn't seem desirable.

cmd/yggdrasil/main.go
```diff
@@ -271,15 +272,53 @@ func main() {
 			n.tun.SetupAdminHandlers(n.admin)
 		}
 	}
+	
+	//Windows service shutdown service
+	minwinsvc.SetOnExit(func() {
+		logger.Infof("Shutting down service ...")
+		sigCh <- os.Interrupt
+		// Wait for all parts to shutdown properly
+		<-doneCh
```
neilalexander, line 281: Is this the part (blocking the exit handler) specifically that fixes the problem? If so, I'm OK with keeping a done channel but then:
1. It doesn't need to be a global, it can just be defined further up in `main`
2. You don't need to replace all of the context code either, you can just do something like this in `main`:
```
done := make(chan struct{})
defer close(done)
```
... and then change `SetOnExit` to:
```
minwinsvc.SetOnExit(func() {
  cancel()
  <-done
})
```
... which would be a far less invasive set of changes.

neilalexander (changes requested):
Think this can be done in a simpler way.

cmd/yggdrasil/main.go
Revertron: Okay, I'll try to simplify it :)

cmd/yggdrasil/main.go
```diff
@@ -53,12 +52,11 @@ func main() {
...
-	// Catch interrupts from the operating system to exit gracefully.
-	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
-
-	// Capture the service being stopped on Windows.
-	minwinsvc.SetOnExit(cancel)
+	
+	done := make(chan struct{})
+	defer close(done)
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
```
neilalexander, line 59: Would probably prefer to keep the `signal.NotifyContext` approach if it achieves the same result, as we're likely to use contexts more in the future.

Revertron: Done.

PR #1157, opened 2024-08-05
Fixed wait for TUN to come up

Revertron:
So, the function waiting for TUN to come up never succeeds:
```
func waitForTUNUp(ch <-chan wgtun.Event) bool {
	t := time.After(time.Second * 5)
	for {
		select {
		case ev := <-ch:
			if ev == wgtun.EventUp {
				return true
			}
		case <-t:
			return false
		}
	}
}
```
I've tried the sleep for one second, and it works flawlessly on several PCs.

Another point - sometimes, if the service stop abruptly (in case of some errors) there is an old hidden device in the system, that we need to uninstall, and then create new.

### DIFF
--- a/src/tun/tun_windows.go
+++ b/src/tun/tun_windows.go
@@ -8,10 +8,12 @@ import (
 	"fmt"
 	"log"
 	"net/netip"
+	"time"
 
 	"github.com/yggdrasil-network/yggdrasil-go/src/config"
 	"golang.org/x/sys/windows"
 
+	"golang.zx2c4.com/wintun"
 	wgtun "golang.zx2c4.com/wireguard/tun"
 	"golang.zx2c4.com/wireguard/windows/elevate"
 	"golang.zx2c4.com/wireguard/windows/tunnel/winipcfg"
@@ -31,14 +33,24 @@ func (tun *TunAdapter) setup(ifname string, addr string, mtu uint64) error {
 		if guid, err = windows.GUIDFromString("{8f59971a-7872-4aa6-b2eb-061fc4e9d0a7}"); err != nil {
 			return err
 		}
-		if iface, err = wgtun.CreateTUNWithRequestedGUID(ifname, &guid, int(mtu)); err != nil {
-			return err
+		tun.log.Printf("Creating TUN")
+		iface, err = wgtun.CreateTUNWithRequestedGUID(ifname, &guid, int(mtu))
+		if err != nil {
+			// Very rare condition, it will purge the old device and create new
+			tun.log.Printf("Error creatung TUN: '%s'", err)
+			wintun.Uninstall()
+			time.Sleep(3 * time.Second)
+			tun.log.Printf("Trying again")
+			iface, err = wgtun.CreateTUNWithRequestedGUID(ifname, &guid, int(mtu))
+			if err != nil {
+				return err
+			}
 		}
-		if !waitForTUNUp(iface.Events()) {
-			return fmt.Errorf("TUN did not come up in time")
-		}
+		tun.log.Printf("Waiting for TUN to come up")
+		time.Sleep(1 * time.Second)
 		tun.iface = iface
 		if addr != "" {
+			tun.log.Printf("Setting up address")
 			if err = tun.setupAddress(addr); err != nil {
 				tun.log.Errorln("Failed to set up TUN address:", err)
 				return err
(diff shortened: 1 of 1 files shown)


Revertron:
BTW, in RiV-mesh they've ditched the waiting for event too. And added a crutch:
```
		for i := 1; i < 10; i++ {
			if err = tun.setupAddress(addr); err != nil {
				tun.log.Errorln("Failed to set up TUN address:", err)
				log.Printf("waiting...")
				if i > 8 {
					return err
				} else {
					time.Sleep(time.Duration(2*i) * time.Second)
				}
			} else {
				break
			}
		}
```

Revertron:
Maybe I'm blind, but I don't even see any events like TunUp in the sources:
https://github.com/WireGuard/wireguard-go/blob/master/tun/tun_windows.go
It's like they just notify on MTU change and that's all.

src/tun/tun_windows.go
```diff
@@ -31,14 +33,24 @@ func (tun *TunAdapter) setup(ifname string, addr string, mtu uint64) error {
 		if guid, err = windows.GUIDFromString("{8f59971a-7872-4aa6-b2eb-061fc4e9d0a7}"); err != nil {
 			return err
 		}
-		if iface, err = wgtun.CreateTUNWithRequestedGUID(ifname, &guid, int(mtu)); err != nil {
-			return err
-		}
-		if !waitForTUNUp(iface.Events()) {
-			return fmt.Errorf("TUN did not come up in time")
+		tun.log.Printf("Creating TUN")
```
neilalexander, line 36: Maybe don't need so many logging lines.

src/tun/tun_windows.go
```diff
@@ -31,14 +33,24 @@ func (tun *TunAdapter) setup(ifname string, addr string, mtu uint64) error {
...
+		iface, err = wgtun.CreateTUNWithRequestedGUID(ifname, &guid, int(mtu))
+		if err != nil {
+			// Very rare condition, it will purge the old device and create new
+			tun.log.Printf("Error creatung TUN: '%s'", err)
```
neilalexander, line 40: Spelling here.

EugeneMartein:
Now I have a two connections on local-link with the same peer.

C:\Users\Admin>yggdrasil -autoconf
2024/08/06 12:07:41 Build name: yggdrasil-develop
2024/08/06 12:07:41 Build version: 0.5.7
2024/08/06 12:07:41 Your public key is 9e4a087d85b69b378556728efa37bdd4b55f9b81a8a342955cdf73bdf1ac9fc8
2024/08/06 12:07:41 Your IPv6 address is 200:c36b:ef04:f492:c990:f553:1ae2:b90
2024/08/06 12:07:41 Your IPv6 subnet is 300:c36b:ef04:f492::/64
2024/08/06 12:07:41 TLS listener started on [fe80::cfb7:d2ea:822c:4fe%Ethernet 2]:40272
2024/08/06 12:07:41 TLS listener started on [fe80::6eba:61cf:4f1b:ba22%Ethernet]:40273
2024/08/06 12:07:41 TCP admin socket listening on 127.0.0.1:9001
2024/08/06 12:07:41 Creating TUN...
2024/08/06 12:07:41 Connected inbound: 200:a9aa:95ca:54f3:41db:8848:f045:8f2b@[fe80::9c4f:d6f0:da8a:9388%Ethernet]:41964, source [fe80::6eba:61cf:4f1b:ba22%Ethernet]:40273
2024/08/06 12:07:41 Using existing driver 0.14
2024/08/06 12:07:41 Creating adapter
2024/08/06 12:07:41 Waiting for TUN to come up
2024/08/06 12:07:42 Setting up address
2024/08/06 12:07:42 TUN is set up successfully
2024/08/06 12:07:50 Connected outbound: 200:a9aa:95ca:54f3:41db:8848:f045:8f2b@[fe80::9c4f:d6f0:da8a:9388%Ethernet]:17537, source [fe80::6eba:61cf:4f1b:ba22%Ethernet]:40312

image

src/tun/tun_windows.go
Revertron: I've removed this line, but I think it is better to leave the others.

src/tun/tun_windows.go
Revertron: Fixed :)

neilalexander (approved):
LGTM!

Issue #1149, opened 2024-06-26
Change multicast code for Android


Revertron:
For fixing multicast in Android app (https://github.com/yggdrasil-network/yggdrasil-android/issues/60, https://github.com/yggdrasil-network/yggdrasil-android/issues/45) we need to change the way `multicast` module gets interface addresses.

One way for this is using `anet` library: https://github.com/wlynxg/anet

But I'm not a go developer... And we need to discuss what way is better.

neilalexander:
Is this the one where Android prevents us from calling `net.Interfaces()` properly? If that's the case, we can just modify the `multicast` module to take a list of interfaces/addresses as input and then have Kotlin pass them in.

Revertron:
Yes, it's the one! :)

> modify the multicast module to take a list of interfaces/addresses as input and then have Kotlin pass them in

Yep, I see this way as better. But how can we implement this? I mean which type of objects will Go part hold|get from Kotlin? Do we need to encode names+IPs in some intermediate format?

bobrofon:
I tried to use [wlynxg/anet](https://github.com/wlynxg/anet) instead of the standard `net` package. `anet` is a little quirky library, but it solves at least part of the problem with multicast.

Here is the patch https://github.com/bobrofon/yggdrasil-go/commit/87234ea2de42512bccf8daaf248d42ed628ced85.
With this patch the multicast module is able to find multicast interfaces and even get the ip addresses. However local peer discovery still doesn't work.
```
I  Initializing multicast 
I  Starting multicast 
I  Starting multicast module
I  Started multicast module
I  Discovered addresses for interface wlan0: [192.168.1.204/24 fd76:dc0f:c24f:0:412b:6731:92c9:ccbf/64 fd76:dc0f:c24f:0:d7a4:c742:869e:f87d/64 fe80::8cf2:ff0:e65:b98e/64]
W  type=1400 audit(0.0:24825): avc: denied { read } for name="somaxconn" dev="proc" ino=1658926 scontext=u:r:untrusted_app:s0:c224,c256,c512,c768 tcontext=u:object_r:proc_net:s0 tclass=file permissive=0 app=eu.neilalexander.yggdrasil
I  Started multicasting on wlan0
I  TLS listener started on [fe80::8cf2:ff0:e65:b98e%wlan0]:37577
[...]

PR #1158, opened 2024-08-06
Restore local peer discovery mechanism on Android 11+

bobrofon:
This solution is bases on https://github.com/wlynxg/anet project. `github.com/wlynxg/anet` is a partial alternative implementation of the `golang.org/x/net` module. The goal of `anet` module is to provide workarounds of the issues https://github.com/golang/go/issues/40569 and https://github.com/golang/go/issues/68082 on Android 11+.

Tested on AOSP 13.

Resolves: #1149

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/kardianos/minwinsvc v1.0.2
 	github.com/quic-go/quic-go v0.45.1
 	github.com/vishvananda/netlink v1.1.0
+	github.com/wlynxg/anet v0.0.4-0.20240806025826-e684438fc7c6
 	golang.org/x/crypto v0.25.0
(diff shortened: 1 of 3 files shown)


Revertron:
@neilalexander are you good with this dependency?

bobrofon:
It looks like https://github.com/bobrofon/yggdrasil-go/commit/cfb329b0793e6fe5729b673a885753d4cc097904 is no longer necessary after the link and multicast fixes were merged into version 0.5.8. At least, I cannot reproduce the livelock on multicasting stop anymore. Removed this commit from the review.

neilalexander (approved):
I checked over the `anet` library and it looks like a nice thin wrapper, so I’m OK with this if you are happy it solves the Android problems.

Issue #1165, opened 2024-08-23
Android build is broken with Go 1.23


bobrofon:
In https://github.com/yggdrasil-network/yggdrasil-go/pull/1158 a new dependency https://github.com/wlynxg/anet was added to the project. A few days later Go 1.23.0 was released. And it was discovered that github.com/wlynxg/anet fails to build with Go 1.23 without an additional ldflag `-checklinkname=0`.
```
$ ./contrib/mobile/build -a
Building aar for Android
gomobile: go build -tags mobile -ldflags -X github.com/yggdrasil-network/yggdrasil-go/src/version.buildName=yggdrasil-go-1.23 -X github.com/yggdrasil-network/yggdrasil-go/src/version.buildVersion=0.5.8-2-g261d8e4 -s -w  -buildmode=c-shared -o=/tmp/gomobile-work-3160141326/android/src/main/jniLibs/x86/libgojni.so ./gobind failed: exit status 1
# gobind/gobind
link: github.com/wlynxg/anet: invalid reference to net.zoneCache
```
More details here: https://github.com/wlynxg/anet/blob/main/README.md#how-to-build-with-go-1230-or-later

PR #1166, opened 2024-08-23
Fix Android build with Go 1.23.0 or later

bobrofon:
The `github.com/wlynxg/anet` library depends on the `//go:linkname` linker feature [1]. However, since Go 1.23.0, the usage of `//go:linkname` has been restricted [2]. And now it's necessary to explicitly specify `-checklinkname=0` linker flag to use it.

[1] https://github.com/wlynxg/anet/blob/main/README.md#how-to-build-with-go-1230-or-later
[2] https://tip.golang.org/doc/go1.23#linker

Resolves: #1165


neilalexander:
Maybe it's better to break out `contrib/mobile` into a separate repository, if having the import is going to cause this kind of breakage...

PR #1175, opened 2024-09-29
Fix ioctl(2) code for OpenBSD

klemensn:
This cleans up the mess to configure an IP address on a tun(4) device.

Handrolling a hardcoded ioctl(2) request is far from perfect, but Go (golang.org/sys/unix) is to blame here.

Tested on OpenBSD 7.6 -current where yggdrasil now drives the interface would use of ifconfig or other helpers.

### DIFF
--- a/src/tun/tun_freebsd.go
+++ b/src/tun/tun_freebsd.go
@@ -1,0 +1,140 @@
+//go:build freebsd
+// +build freebsd
+
+package tun
+
+import (
+	"encoding/binary"
+	"fmt"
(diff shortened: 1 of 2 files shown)


klemensn:
Rebased onto develop as of now.

neilalexander (approved):
Looks good, thanks!

klemensn:
Sadly, I just noticed you squashed all commits, thereby omitting all their messages, i.e. insights into what was wrong, from the commit history.

PR #1186, opened 2024-10-08
Use UNIX socket patch from url struct

klemensn:
No need to extract it again when the url package provides it for us:
```
$ jq -n '{"AdminListen":"unix:///tmp/ygg.sock"}' | ./yggdrasil -useconf | grep 'admin socket'
2024/10/08 22:41:11 UNIX admin socket listening on /tmp/ygg.sock
```

Follow-up on #1176

### DIFF
--- a/src/admin/admin.go
+++ b/src/admin/admin.go
@@ -238,28 +238,27 @@ func (a *AdminSocket) listen() {
 	if err == nil {
 		switch strings.ToLower(u.Scheme) {
 		case "unix":
-			file := strings.TrimPrefix(listenaddr, "unix://")
-			if _, err := os.Stat(file); err == nil {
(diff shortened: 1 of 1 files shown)


neilalexander (commented):
Looks good at first glance but can you please just check this does the right thing when there are complex characters in the file path that may or may not be URL-encoded?

klemensn:
> Looks good at first glance but can you please just check this does the right thing when there are complex characters in the file path that may or may not be URL-encoded?

It now decodes URL encoded file paths, e.g. `unix:///tmp/y%40gg.sock` becomes `/tmp/y%40gg.sock` / `/tmp/y@gg.sock` with master / this PR, respectively.

Is that something to realistically

PR #1190, opened 2024-10-13
Dynamically linked libraries with docker support

HappyHakunaMatata:
Added dynamic library generation, so now it's possible to compile the library and use it in your .NET project.
Flags have been replaced with Cobra commands (Docker launch has been maintained).
Added a database to store all data retrieved from the API.


neilalexander (commented):
This is an awful lot of code and, aside from the functions in `src/core`, I'm not sure that any of it needs to be in the mainline Yggdrasil distribution. Especially database code, for which it's not clear to me what its purpose is.

HappyHakunaMatata:
> This is an awful lot of code and, aside from the functions in `src/core`, I'm not sure that any of it needs to be in the mainline Yggdrasil distribution. Especially database code, for which it's not clear to me what its purpose is.

As I see it, it's currently impossible to save information from the API such as Peer info, etc.(Now it’s only possible to monitor). Moreover, after the session ends, this information is not available in yggdrasilctl, so the database is used to store the data. Dynamic library compilation can be useful if you want to use Yggdrasil in .NET (for example in your ASP NET project you just need to create a custom marshaler). In my opinion Cobra command fits better than Flags.

PR #1193, opened 2024-10-20
genkeys, yggdrasilctl: Use pledge(2) on OpenBSD

klemensn:
Restrict system operations of CLI tools with https://man.openbsd.org/pledge.2.

https://pkg.go.dev/suah.dev/protect abstracts the OS specific code, i.e. is a NOOP on non-OpenBSD systems.

This PR is to gauge upstream interest in this direction;  my OpenBSD port of yggdrasil already pledges the daemon,
resulting in minimal runtime privileges, but there are still a few rough edges:
https://github.com/jasperla/openbsd-wip/blob/master/net/yggdrasil/patches/patch-cmd_yggdrasil_main_go#L80


klemensn:
Rebased onto develop.

@neilalexander any thoughts on the general direction of adding pledge and unveil to yggdasil?

It would be code effectively running only on OpenBSD, but it being there should help developers on other OSes applying similar security mechanisms, e.g. Linux/seccomp or FreeBSD/capsicum.

neilalexander:
In principle I'm not opposed but I am a little bit concerned about absorbing more and more platform-specific code for platforms that neither I nor Arceliar run or can easily test. If you're volunteering to maintain OpenBSD-specific code on an on-going basis, that's a different story.

klemensn:
> In principle I'm not opposed but I am a little bit concerned about absorbing more and more platform-specific code for platforms that neither I nor Arceliar run or can easily test.

That's expected, hence my asking before investing my time and yours upstream, only to keep work as local patches after all because of stuff like this.

> If you're volunteering to maintain OpenBSD-specific code on an on-going basis, that's a different story.

Sure, I'll gladly put the had on for any OpenBSD-specific bits in yggdrasil, if that helps.

I've already ported it and am actively maintaining it;  it still lives in jasperla/openbsd-wip until I'm confident enough that everything will work as expected for OpenBSD users, at which point I'll send my port out for review and import it into openbsd/ports proper.

All my PRs so far came out of that, making stuff "just work" on my nodes.

klemensn:
> I've already ported it and am actively maintaining it; it still lives in jasperla/openbsd-wip until I'm confident enough that everything will work as expected for OpenBSD users, at which point I'll send my port out for review and import it into openbsd/ports proper.

I missed that we already had a heavily outdated [net/yggdrasil-go](https://github.com/openbsd/ports/tree/master/net/yggdrasil-go) port at 0.5.3, so I just polished and updated that and took maintainership to take care of it in the future.

klemensn:
> my OpenBSD port of yggdrasil already pledges the daemon, resulting in minimal runtime privileges, but there are still a few rough edges

Such as https://github.com/yggdrasil-network/yggdrasil-go/pull/927/files#r1826606055

The lack of serialisation between goroutines makes them race against each other.
The main goroutines wants to drop privileges, but has no way of knowing whether all privileged tasks on startup have been performed already.

klemensn:
@neilalexander Is this something you'd consider merging upstream?

If not, I'll carry local patches in the OpenBSD port and drop these PRs.

If so, then perhaps merge this one and refine, i.e. tighten security, in follow-up ones?

neilalexander:
Happy to merge this for now, yep.

PR #1194, opened 2024-10-22
Use unveil(2) on OpenBSD

klemensn:
After #1175 removed ioctl(2) fallback code shelling out to ifconfig(8), there is no code left (compiled on OpenBSD) that would fork(2) or execve(2).

Drop the ability to run any executable file to double down on this, thus reducing the attack surface of this this experimental, internet facing daemon running as root.

pledge(2) is doable, but needs more polish.
unveil(2), however, is as simple as it gets.

On other systems, this code is a NOOP, but can still help to implement similar safety belts.


klemensn:
Like #1193, this adds https://pkg.go.dev/suah.dev/protect to `go.{mod,sum}`, so iff either of the two PRs get merged, the other needs updating.

klemensn:
Thoughts?

I've been running this together with pledge and privilege drop to a dedicated user for weeks now without issues.

PR #1201, opened 2024-11-03
Create admin socket synchronously before privdrop

klemensn:
Creating UNIX sockets the listen() goroutine that races against the main
one dropping to an unprivileged user may cause startup failure when
privdrop happens before privileged filesystem access.

Setup or fail in New() and only do listen(2) in listen() to avoid this.

```
# yggdrasil -autoconf -user nobody
2024/11/03 21:15:27 Build name: yggdrasil-go
2024/11/03 21:15:27 Build version: 0.5.9
...
2024/11/03 21:15:27 Admin socket failed to listen: listen unix /var/run/yggdrasil.sock: bind: permission denied
```

Rerun, now the order is flipped:
```
# yggdrasil -autoconf -user nobody
2024/11/03 21:15:34 Build name: yggdrasil-go
2024/11/03 21:15:34 Build version: 0.5.9
[...]
2024/11/03 21:15:34 UNIX admin socket listening on /var/run/yggdrasil.sock
[...]
```

Fixes #927.

### DIFF
--- a/src/admin/admin.go
+++ b/src/admin/admin.go
@@ -83,6 +83,52 @@ func New(c *core.Core, log core.Logger, opts ...SetupOption) (*AdminSocket, erro
 	if a.config.listenaddr == "none" || a.config.listenaddr == "" {
 		return nil, nil
 	}
+
+	listenaddr := string(a.config.listenaddr)
+	u, err := url.Parse(listenaddr)
+	if err == nil {
+		switch strings.ToLower(u.Scheme) {
+		case "unix":
+			if _, err := os.Stat(u.Path); err == nil {
+				a.log.Debugln("Admin socket", u.Path, "already exists, trying to clean up")
+				if _, err := net.DialTimeout("unix", u.Path, time.Second*2); err == nil || err.(net.Error).Timeout() {
+					a.log.Errorln("Admin socket", u.Path, "already exists and is in use by another process")
+					os.Exit(1)
+				} else {
+					if err := os.Remove(u.Path); err == nil {
+						a.log.Debugln(u.Path, "was cleaned up")
+					} else {
+						a.log.Errorln(u.Path, "already exists and was not cleaned up:", err)
+						os.Exit(1)
+					}
+				}
+			}
+			a.listener, err = net.Listen("unix", u.Path)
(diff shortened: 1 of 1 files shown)


klemensn:
In a single-core VM the timing is different enough for above mentioned `-autoconf -user nobody` to never trigger, but disabling TUN and multicast, i.e. disabling extra work in between, makes it always fail:

```
# uname -a
Linux alpine 6.6.59-0-virt #1-Alpine SMP PREEMPT_DYNAMIC 2024-11-01 11:02:13 x86_64 Linux
# nproc
1
# yggdrasil -version
Build name: yggdrasil
Build version: 0.5.9
```

```
# yggdrasil -genconf -json | jq '.MulticastInterfaces = [] | .IfName = "none"' | yggdrasil -useconf -user nobody
2024/11/03 22:46:54 Build name: yggdrasil
2024/11/03 22:46:54 Build version: 0.5.9
2024/11/03 22:46:54 Your public key is 52f2d9aa58335096fcefc78440230d6d51ac84a14df5064f1b5f9ab293e58265
2024/11/03 22:46:54 Your IPv6 address is 201:b434:9956:9f32:bda4:c40:e1ee:ff73
2024/11/03 22:46:54 Your IPv6 subnet is 301:b434:9956:9f32::/64
2024/11/03 22:46:54 Admin socket failed to listen: listen unix /var/run/yggdrasil.sock: bind: permission denied
```

klemensn:
Anyone?

Introducing this synchronisation point allows me to also reliably drop privileges further via pledge(2) on OpenBSD, since it is then guaranteed that the socket is set up already, see https://github.com/yggdrasil-network/yggdrasil-go/pull/1193#issuecomment-2453058057

neilalexander:
Not ignoring this, I just haven't had the time for the last week to review PRs due to work and other things. I'll try to take a look later today.

neilalexander (commented):
I'm wondering if the better approach might just be to move some of the initialisation logic out of the `listen()` goroutine and into `New()`, such that by the time the function returns, we know if it succeeded or failed and the listener would already be created?

neilalexander (approved):
Looks good, thanks!

PR #1203, opened 2024-11-04
Rewrite chuser() for simplicity and correctness

klemensn:
- Use unambiguous variable names (w/o package name conflict).
- Fail on invalid input such as the empty string or `:`.
- Do not change group without user, i.e. fail on `:group`.
- Parse input using mnemonic APIs.
- Do not juggle between integer types.
- Unset supplementary groups.
- Use set[ug]id(2) to follow the idiom of OpenBSD base programs.
  (cannot use setres[ug]id(2) as macOS does not have them.)

Includes/Supersedes #1202.
Fixes #927.

I only tested on OpenBSD (so far), but other systems should just work.


neilalexander:
I'm OK with this, we might be able to write a unit test in `core_test.go` to prove that it does the right thing on Linux etc.

klemensn:
> I'm OK with this, we might be able to write a unit test in `core_test.go` to prove that it does the right thing on Linux etc.

I simply replaced the existing code.

Looking into tests now...

klemensn:
Now with macOS building, still looks good on OpenBSD:
```
$ id nobody
uid=32767(nobody) gid=32767(nobody) groups=32767(nobody)
$ ps -c -U nobody -o command,user,ruser,svuid,group,rgroup,svgid,supgrp
COMMAND          USER     RUSER    SVUID GROUP    RGROUP   SVGID SUPGRP
yggdrasil        nobody   nobody   32767 nobody   nobody   32767 nobody
```

klemensn:
> I'm OK with this, we might be able to write a unit test in `core_test.go` to prove that it does the right thing on Linux etc.

`test_chuser_unix.go` testing `chuser()` alone is a start.

klemensn:
Unless you run tests as root (who does that?), there is no way to test all of that code automatically.

@neilalexander I've abused an optional function parameter to cover the actual set*id(2) syscalls in CI.
It should work, but it is more of a PoC, which is to say, that I'd rather not have that `chuser()` change land.

klemensn:
> Unless you run tests as root (who does that?), there is no way to test all of that code automatically.
> 
> @neilalexander I've abused an optional function parameter to cover the actual set*id(2) syscalls in CI. It should work, but it is more of a PoC, which is to say, that I'd rather not have that `chuser()` change land.

https://github.com/yggdrasil-network/yggdrasil-go/actions/runs/11787911199/job/32834078977?pr=1203#step:5:19

Linux drops privileges fine in CI.

neilalexander:
Is this ready for review in that case?

klemensn:
Yes, the PR is ready as-is.

neilalexander (approved):
LGTM, thanks!

PR #1215, opened 2024-12-12
Use pledge(2) on OpenBSD

klemensn:
Straight forward thanks to all privileged operations being done early enough during startup.

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -312,6 +312,21 @@ func main() {
 		}
 	}
 
+	// Promise final modes of operation.  At this point, if at all:
+	// - raw socket is created/open
+	// - admin socket is created/open
+	// - privileges are dropped to non-root user
+	//
+	// Peers, InterfacePeers, Listen can be UNIX sockets;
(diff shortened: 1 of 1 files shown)


klemensn:
I've been running this for months now on nodes with and without local multicast discovery.

Switching the control socket from UNIX to TCP also does the right thing.

cmd/yggdrasil/main.go
```diff
@@ -312,6 +312,22 @@ func main() {
 		}
 	}
 
+	// Promise final modes of operation.  At this point, if at all:
+	// - raw socket is created/open
+	// - admin socket is created/open
+	// - privileges are dropped to non-root user
+	promises := []string{"stdio", "inet", "dns"}
+	if strings.HasPrefix(cfg.AdminListen, "unix://") {
```
neilalexander, line 320: What about `Peers`, `InterfacePeers` or `Listen`? Those can be `unix://` too.

klemensn: Right, I'll fix that and update the PR, thanks.

PR #1223, opened 2025-01-03
Introduce VSOCK link type

nagy:
This address family is used in communication between virtual machines and their hosts. Advantages include that no virtual ethernet adapter and their respective address configuration and routing need to be setup. Rather, with this new link type, only a single yggdrasil interface can exist inside of the virtual machine.

It can also be used inside of containers. There, the advantage over existing link types like unix sockets include, that no mount point need to be shared with the host and container. This provides more isolation.

More information:

https://man7.org/linux/man-pages/man7/vsock.7.html

https://gist.github.com/nrdmn/7971be650919b112343b1cb2757a3fe6


nagy:
Would you be open to external link types for example via a plugin system? For example if you would have an executable program in `/usr/bin/yggdrasil-transport-<name>` or more likely somewhere in the `PATH` then the configuration URL like `vsock://local:1234` would be passed to it as a first argument to the program `yggdrasil-transport-vsock`, if it exists. The actual communication happens via stdio. This way we would not need to maintain this dependency in the main repository. People could install new transports via packages via their package managers and we could experiment with new transports in other languages.


> , given they mostly just encourage people to do silly things topology-wise.

Of course you would know more about these topology problems than I do, but I want to say that people sometimes rely on these exotic transports to bypass firewalls.

state-plumber:
The most generalised way of doing this that I can think of would be support peering over [UCSPI](https://cr.yp.to/proto/ucspi.txt).
- listeners are specified as a list of open file-descriptors where `accept` will return new connections
  `{ Listen: [ fd://0 ] }`
- peers are specified as pairs of open R/W file-descriptors
  `{ Peers: [ fd://6/7 ] }`

Yggdrasil is started as "[vsockserver-socketbinder](https://spectrum-os.org/git/ucspi-vsock/tree/vsockserver-socketbinder.c) CID PORT yggdrasil" or as "[vsockclient](https://spectrum-os.org/git/ucspi-vsock/tree/vsockclient.c) CID PORT yggdrasil". Multiple listeners and clients may be specified by [moving](https://skarnet.org/software/execline/fdmove.html) file-descriptors around before exec'ing yggdrasil.

majestrate:
> The most generalised way of doing this that I can think of would be support peering over [UCSPI](https://cr.yp.to/proto/ucspi.txt).
> - listeners are specified as a list of open file-descriptors where `accept` will return new connections
>   `{ Listen: [ fd://0 ] }`

I'd assume there is a constraint where the underlying mechanism which produces the fd needs to come from a call to `socket()`, correct? (implementation detail, yes. just curious ) 

> - peers are specified as pairs of open R/W file-descriptors
>   `{ Peers: [ fd://6/7 ] }`
> 
> Yggdrasil is started as "[vsockserver-socketbinder](https://spectrum-os.org/git/ucspi-vsock/tree/vsockserver-socketbinder.c) CID PORT yggdrasil" or as "[vsockclient](https://spectrum-os.org/git/ucspi-vsock/tree/vsockclient.c) CID PORT yggdrasil". Multiple listeners and clients may be specified by [moving](https://skarnet.org/software/execline/fdmove.html) file-descriptors around before exec'ing yggdrasil.
> 


this would work and would let us gut support for link dialects we do not want to have to support, and externalize the implementation to people who want to focus on that. main gotcha is documentation. djbware is notoriously obtuse to use for end users born in the 21st century.

PR #1243, opened 2025-04-11
[deb] Don't run as root

ledlamp:
Any reason it needs to run as root? Seems to be working fine running as a user because it has the capabilities it needs.


l29ah:
Admin socket failed to listen: listen unix /var/run/yggdrasil.sock: bind: permission denied

ledlamp:
Add RuntimeDirectory=yggdrasil and listen on  /run/yggdrasil/yggdrasil.sock

PR #1253, opened 2025-05-02
Fix default config permissions

mirefly42:
Currently, all init scripts, except for systemd, will generate a config file with default permissions, which is usually `rw-r--r--`.
This is bad, because the config contains a private key.

The systemd service does `chmod 640` after creating the config, which is much better than just leaving it readable for everyone forever, but there is still a slight chance that some malicious program might steal the private key during the time window between key creation and chmod.

For this reason, in this pull request I use `umask 037`, so the config won't have read permission for others in the first place.

Note that I have only tested openrc and systemd services.

Also, I'm not sure what to do with the contrib/msi/build-msi.sh script, which creates a bat file that generates a config. I don't know anything about file permissions on windows, however, it seems that the bat file generates the config into a user's personal directory, so maybe it's already somewhat fine.

### DIFF
--- a/contrib/busybox-init/S42yggdrasil
+++ b/contrib/busybox-init/S42yggdrasil
@@ -3,7 +3,7 @@
 CONFFILE="/etc/yggdrasil.conf"
 
 genconf() {
-	/usr/bin/yggdrasil -genconf > "$1"
+	(umask 037 && /usr/bin/yggdrasil -genconf > "$1")
 	return $?
 }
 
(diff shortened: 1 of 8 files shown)


mirefly42:
Now I'm realizing that this is a bit of duplicated code across all the different init systems and install scripts, and it is very easy for someone to forget about this problem when creating new scripts.
I have a couple ideas for a better solution.

* Introduce a command line argument, something like `-genconffile` (similar to `-useconffile`), which would create a file with right permissions by itself.
* When using `-genconf`, check the permissions of stdout using `fstat`, and emit a warning if the permissions are insecure. This technique is used in wireguard-tools.

What does everyone think about this?

PR #1256, opened 2025-05-05
Add option to log to stderr

ehmry:
UNIX provides file-descriptor 2 for diagnostic output.

https://pubs.opengroup.org/onlinepubs/9799919799/


ehmry:
Whatever, this is a dumb option for a dumb problem.

PR #1275, opened 2025-08-11
OpenBSD: Pledge full filesystem read for Go's resolv.conf polling

klemensn:
sys/kern/kern_pledge.c r1.329[0] removed the unveil bypass for "dns",
so "rpath" is needed for Go's DNS to stat(2) it.

Since current "/ rwc" and "cpath" with the new "rpath" amount to full
read access, there is no point in unveiling anymore.

0: https://github.com/openbsd/src/commit/8d49ad01ac8af306145a19ff67bb1f10cd3ef57f

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -41,20 +41,6 @@ type node struct {
 
 // The main function is responsible for configuring and starting Yggdrasil.
 func main() {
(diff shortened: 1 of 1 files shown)


klemensn:
See unveil #1194 and pledge #1215 additions for context.

neilalexander (approved):

PR #1278, opened 2025-09-01
Add multi-arch Docker build & GHCR publishing

Forne:
# Summary
This PR addresses failures to run Yggdrasil on ARM systems. The root cause was the lack of ARM artifacts/images, which led to exec format error and similar issues.

## What’s added:
- ```Dockerfile.multiarch``` — multi-stage Go build that correctly propagates GOOS/GOARCH for linux/amd64, linux/arm64, linux/armhf and linux/armel platform.
- ```entrypoint.sh``` - Introduced ENV **ALLOW_IPV6_FORWARDING**. When set to a truthy value (e.g., true), the entrypoint executes: ```sysctl -w net.ipv6.conf.all.forwarding=1```.
- GitHub Action for multi-arch builds and publishing to GHCR — triggered via ```workflow_dispatch```, push to ```master``` and release via tags (with docker semantic tags e.g. v0.5.12 → 0.5.12, 0.5, 0).

Example published images:
[https://github.com/Forne/yggdrasil-go/pkgs/container/yggdrasil-go](https://github.com/Forne/yggdrasil-go/pkgs/container/yggdrasil-go)

## Testing
✅ Ubuntu (24.04, amd64) — image runs correctly.
✅ macOS (Apple Silicon, arm64) — image runs correctly.
✅ MikroTik RouterOS (arm64) — image runs under the RouterOS container package.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,6 +2,9 @@ name: Docker Build
 
 on:
   workflow_dispatch:
+  push:
+    branches: [ main, master ]
+    tags: [ 'v*' ]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
Looks good, thanks!

PR #1290, opened 2025-11-29
Update yggdrasil.service.debian

Revertron:
Can't start service without reading config.
Without explicit allow directive it can't read it.


majestrate:
LGTM

PR #1336, opened 2026-04-04
Change ownership of admin socket before dropping permissions

state-plumber:
This is to allow access to the socket by members of the group that permissions are dropped to.

### DIFF
--- a/cmd/yggdrasil/chuser_unix.go
+++ b/cmd/yggdrasil/chuser_unix.go
@@ -4,6 +4,8 @@ package main
 
 import (
 	"fmt"
+	"net/url"
+	"os"
 	"os/user"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


majestrate (approved):
Looks fine

cmd/yggdrasil/chuser_unix.go
```diff
@@ -48,6 +50,16 @@ func chuser(input string) error {
 		gid, _ = strconv.Atoi(usr.Gid)
 	}
 
+	if adminSockUrl != "" {
+		u, err := url.Parse(adminSockUrl)
+		if err == nil && u.Scheme == "unix" {
+			err = os.Chown(u.Path, uid, gid)
+		}
+		if err != nil {
```
neilalexander, line 58: Do we intentionally want to fall through to this on the `url.Parse` returning an error case?

state-plumber: Yes, it's an intentional fallthrough so there is only one early return check in this block. It's not exactly idiomatic.

neilalexander (approved):
LGTM, thanks!

PR #1337, opened 2026-04-04
Add -notifyfd flag for s6-style readiness notification

state-plumber:
Write a newline to the specified file-descriptor to signal that yggdrasil is fully initialised and ready.

s6 service startup notifications are described [here](https://skarnet.org/software/s6/notifywhenup.html).

Related: #1148

### DIFF
--- a/cmd/yggdrasil/main.go
+++ b/cmd/yggdrasil/main.go
@@ -55,6 +55,7 @@ func main() {
 	getpkey := flag.Bool("publickey", false, "use in combination with either -useconf or -useconffile, outputs your public key")
 	loglevel := flag.String("loglevel", "info", "loglevel to enable")
 	chuserto := flag.String("user", "", "user (and, optionally, group) to set UID/GID to")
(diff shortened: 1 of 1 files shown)


majestrate (commented):
Maybe It would be better to use an env var instead of a cli opt ?

state-plumber:
> Maybe It would be better to use an env var instead of a cli opt ?

I have only seen this option as a command-line flag.

state-plumber:
The next release of s6 will have a `s6-notify-socket-from-fd` shim that can chain-load yggdrasil and notify systemd using this `-notifyfd` feature. That means this change is sufficient to support both s6 and systemd notification.

majestrate:
this PR looks fine to me, I'll wait for @neilalexander to approve.

neilalexander:
Looks like the linter is unhappy, if you don't mind taking a look.

state-plumber:
I might have fixed it?

neilalexander (approved):
Looks good, thanks!

PR #1340, opened 2026-04-10
fix: reject malformed network input in parsers

AlexMelanFromRingo:
## Summary
- validate handshake metadata field lengths before fixed-size reads
- reject truncated multicast advertisements before slicing the hash payload
- add regression tests for malformed and truncated input

## Why
Both parsers currently trust length information from the incoming payload a bit too much. Malformed network input can reach fixed-size reads/slices and panic the process instead of being rejected cleanly.

## Testing
- go test ./...


majestrate:
LGTM

neilalexander (approved):
LGTM, thanks!

PR #1342, opened 2026-05-06
Allow WebSocket listeners to configure browser origins

asciimoth:
Adds an `origin` query option for `ws://` listener URLs so peer operators can
explicitly allow browser-hosted WebSocket clients.

- `ws://host:port` keeps the existing same-origin behavior
- `ws://host:port?origin=demo.example.org` allows that origin host
- `ws://host:port?origin=https://demo.example.org` allows that scheme and host
- repeated `origin=` parameters allow multiple origin patterns
- `origin=*` intentionally disables origin verification for public WebSocket
  peer endpoints

## Problem
I've implemented a WASM based browser demo yggdrasil node to found that it
cannot directly dial any existing public `ws://` or `wss://` peers.

Browsers always include an `Origin` header in WebSocket handshakes, and the
JavaScript `WebSocket()` constructor does not allow applications to override or
remove arbitrary handshake headers.  
This means a browser demo served from an origin such as
`http://127.0.0.1:8000` cannot connect to a public peer whose WebSocket server
only accepts same-origin handshakes.

### DIFF
--- a/src/core/link_ws.go
+++ b/src/core/link_ws.go
@@ -30,8 +30,9 @@ type linkWSListener struct {
 }
 
 type wsServer struct {
-	ch  chan *linkWSConn
-	ctx context.Context
+	ch            chan *linkWSConn
+	ctx           context.Context
+	acceptOptions *websocket.AcceptOptions
 }
 
 func (l *linkWSListener) Accept() (net.Conn, error) {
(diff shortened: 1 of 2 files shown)


majestrate:
Changeset seems reasonable.

asciimoth:
@neilalexander could you review pls

neilalexander (approved):
LGTM

PR #1343, opened 2026-05-07
fix: avoid panics on edge-case input across modules

AlexMelanFromRingo:
## Summary
- ipv6rwc: validate IPv6 packet length before reading the version nibble in writePC
- config: guard the BOM check against configs shorter than two bytes
- admin: replace unchecked net.Error type assertion with errors.As; tolerate empty unix socket paths
- multicast: log and continue on ReadFrom errors instead of panicking; use checked type assertion on UDPAddr
- mobile: reject negative length in SendBuffer; nil-check AddrForKey in GetPeersJSON and SummaryForConfig
- admin/get{tree,paths,sessions}: skip entries when AddrForKey returns nil instead of dereferencing
- core/nodeinfo: validate the requested public key length in nodeInfoAdminHandler, matching the other proto handlers
- add regression tests for the panic paths

## Why
A handful of error paths and platform-API edge cases reach fixed-size
indexing or unchecked type assertions before any length validation.
Most are reachable only locally (an empty config piped to -useconf,
a 0-byte packet from the mobile bindings, an admin DialTimeout error
that doesn't satisfy net.Error on some platforms), but they crash the
daemon hard. Have them return errors or skip the entry instead.

## Testing
- go test ./...
- go vet ./...

### DIFF
--- a/src/ipv6rwc/ipv6rwc.go
+++ b/src/ipv6rwc/ipv6rwc.go
@@ -281,12 +281,15 @@ func (k *keyStore) readPC(p []byte) (int, error) {
 }
 
 func (k *keyStore) writePC(bs []byte) (int, error) {
-	if len(bs) < 40 {
-		strErr := fmt.Sprint("undersized IPv6 packet, length: ", len(bs))
-		return 0, errors.New(strErr)
+	if len(bs) == 0 {
+		return 0, errors.New("empty packet")
 	}
 	if bs[0]&0xf0 != 0x60 {
 		return 0, errors.New("not an IPv6 packet") // not IPv6
+	}
+	if len(bs) < 40 {
+		strErr := fmt.Sprint("undersized IPv6 packet, length: ", len(bs))
+		return 0, errors.New(strErr)
 	}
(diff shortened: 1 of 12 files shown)


src/ipv6rwc/ipv6rwc.go
```diff
@@ -281,13 +281,13 @@ func (k *keyStore) readPC(p []byte) (int, error) {
 }
 
 func (k *keyStore) writePC(bs []byte) (int, error) {
-	if bs[0]&0xf0 != 0x60 {
-		return 0, errors.New("not an IPv6 packet") // not IPv6
-	}
 	if len(bs) < 40 {
```
majestrate: in the case of ipv4 packets this would give an incorrect error message.

AlexMelanFromRingo: good catch, reordered the checks so non-IPv6 first bytes get "not an IPv6 packet" before the length error fires. pushed in 9f21bb6.

majestrate (approved):
looks fine

neilalexander (approved):
LGTM, thanks!

PR #1362, opened 2026-07-30
Add experimental Plan 9 / 9front support

kiljoy001:
## Summary

Adds the platform-specific layers to run yggdrasil as a **full node on Plan 9 (9front)**: a TUN adapter, link-local multicast peer discovery, and platform defaults. No changes to the supported-platform code paths — the shared multicast logic (`_announce`/`listen`/`_stop`) is byte-for-byte unchanged.

## What was verified

- **Builds**: `plan9/amd64`, `plan9/386`, `linux`, `darwin`, `freebsd`, `openbsd` all `go build` clean; `go vet` clean on plan9; `go test ./src/...` passes on linux (no regression).
- **Runtime on a real 9front grid**: the cross-compiled binary runs as a node — the `pkt` TUN interface comes up with the `0200::/7` address (MTU clamped to the medium's 4096), multicast beacons are sent and received via `/net/udp`, and peers connect **both inbound and outbound**. Graceful shutdown on the `interrupt` note unbinds the interface cleanly.

## Design

**TUN** (`src/tun/tun_plan9.go`): Plan 9 has no in-kernel TUN device. The equivalent is a `pkt` packet interface (`ipifc`) per `ip(3)` — `bind pkt`, `add` the `0200::/7` address, and read/write raw IPv6 through `/net/ipifc/N/data`. This is the same primitive `6in4(8)`/`ppp(8)` use. Implemented as a `wgtun.Device` so the existing TUN read/write goroutines work unchanged. On startup it reclaims any stale ipifc still carrying our address (Plan 9 doesn't reap a `pkt` ipifc when its owner is hard-killed, unlike a Linux tun).

**Multicast** (`src/multicast/multicast_plan9.go`): `golang.org/x/net/ipv6` is stubbed on Plan 9, so the beacon socket is driven natively through `/net/udp` in "headers" mode (`announce *!port`, `addmulti` to join `ff02::114`, 52-byte `Udphdr` with big-endian ports). To avoid touching the shared multicast code, `plan9Sock` mirrors the exact subset of `*ipv6.PacketConn` method signatures that `multicast.go` calls (`JoinGroup`/`WriteTo`/`ReadFrom`/`Close`); `ReadFrom` returns an `*ipv6.ControlMessage{Dst:…}` so the shared `listen()`'s existing `rcm.Dst` checks work as-is. Only the `Multicast` struct + `_start` (socket creation) are per-platform — the non-Plan-9 versions are moved **verbatim** to `multicast_sock.go`. Reclaims stale UDP convs on startup + retries `announce` to handle a port freed mid-shutdown.

## Caveats / notes

- **Admin socket**: listens on TCP loopback (`tcp://localhost:9001`) and requires a configured loopback interface (`ipconfig loopback /dev/null 127.1 /128`). The admin socket has no auth and relies on loopback binding for security, so it intentionally does not fall back to a wildcard bind. Documented in the README.
- **Hard shutdown**: a `slay`/crash leaves the `pkt` ipifc and `0200::/7` route behind; the next start reclaims them. A clean `interrupt`-note shutdown unbinds normally.
- `contrib/ansible/genkeys.go` is excluded from plan9 builds (the `cheggaaa/pb` `termutil.TerminalSize` symbol has no plan9 definition — appears to be an upstream `pb` bug worth filing separately).
- CI adds plan9/amd64 + plan9/386 **cross-compile** jobs (no plan9 test runner; tests can't run, only build-verified).
- Runtime-verified on 9front **amd64**; the 386 target compiles but wasn't run.

This is experimental and Plan 9 is a niche target that the maintainers can't easily test, so I'm happy to maintain this going forward and to open a tracking issue if useful. Feedback welcome — in particular on whether the `multicast_sock.go` split (moving the struct + `_start` out of `multicast.go`) is an acceptable way to branch the plan9 execution, or if you'd prefer a different structure.


state-plumber:
**kiljoy001** has a history of spamming the Plan9 communtity with slop: https://9fans.topicbox.com/groups/9fans/T1338384c5eaf40a5-M3652b83598097194662aea16/new-9p-microkernel

kiljoy001:
If you call a single post spam, god help you.
Screenshot from 2026-08-22 15-57-16
Screenshot from 2026-08-22 16-08-11

Anyway, here is a screenshot of it working. The peer injection is not part of the pr.

